Issue #18, opened 2023-09-29
FEATURE - User Bugs / Feature Requests from `SettingsView`


mikaelacaron:
## Describe the feature you want the app to have that it currently doesn't
* Create a new row under the GitHub Repo row in the 1st section of `SettingsView` called "Request a New Feature" that links the user to create a new feature in the GitHub repo
   * Choose any SF symbol that fits for this "request a new feature" row
   * Link to here:`https://github.com/mikaelacaron/Basic-Car-Maintenance/issues/new?assignees=&labels=feature+request&projects=&template=feature-request.md&title=FEATURE+-`
* Create a new for row for "Report a Bug"
   * Use the `ladybug` SF symbol
   * Link here: `https://github.com/mikaelacaron/Basic-Car-Maintenance/issues/new?assignees=&labels=bug&projects=&template=bug-report.md&title=BUG+-`

## Describe alternatives you've considered
* Yes this requires someone to have a GitHub account to actually find this useful, but that's fine for now

## Additional context
* Add any other context or screenshots about the feature request here.

mishapark:
Hi, please assign this issue to me

mikaelacaron:
> Hi, please assign this issue to me

@mishapark you can only work on **ONE** issue at a time, please go through and delete your comments on all the other issues, you commented on already. Do NOT start working on this until Oct 1, because the code base isn't ready yet

mishapark:
Got it, thanks for letting me know

mikaelacaron:
Closed with #21

PR #21, opened 2023-10-01
Added two rows (new feature/report bug) to SettingsView

mishapark:
# What it Does
* Closes issue #18 
* Added a row for GitHub users to submit new feature requests, and a row to report bugs  in SettingsView

# How I Tested
For example:
* Run the application
* Tap on all three rows including GitHub Repo
* All work as expected, taking the user to Safari to an appropriate link

# Notes
* Haven't added my team ID, as I am still battling with Apple to enroll into the dev program

# Screenshots

### DIFF
--- a/Basic-Car-Maintenance/Shared/Localizable.xcstrings
+++ b/Basic-Car-Maintenance/Shared/Localizable.xcstrings
@@ -46,6 +46,12 @@
     "Profile" : {
 
     },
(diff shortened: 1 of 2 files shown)


mikaelacaron:
@mishapark Great work!
The team ID is only needed when you run on a real device, and it's *not* committed to git, so no one else will see it

Issue #24, opened 2023-10-01
BUG - Vehicle with empty details getting saved


ayushs-2k4:
## Describe the bug
* Vehicle getting added without even filling the details.

## To Reproduce
1. Go to 'Settings'
2. Tap on the button 'Add Vehicle'
3. Do not fill in any details
4. Tap on 'Add' Button
5. Empty details Vehicle will be saved

## Expected behavior
* Vehicle should not get added.

## Screenshots
![Simulator Screenshot - iPhone 15 Pro - 2023-10-01 at 19 17 52](https://github.com/mikaelacaron/Basic-Car-Maintenance/assets/63922782/4a245328-aa69-4b80-a204-e464c5c6abff)

## Device Info (please complete the following information):
 * Device: Phone14 Pro
 * OS: iOS 17.0
 * App Version 1.0(1)

## Additional Context
* N/A

ayushs-2k4:
I am working on this issue.

mikaelacaron:
@ayushs-2k4 Good catch!

Be sure to fill in every section of the template, this section was left blank /  the default template
> Describe the bug
A clear and concise description of what the bug is.

ayushs-2k4:
Oh, it slipped from my eyes to write description of bug. BTW, I have submitted a PR fixing the bug by adding a check before saving a new Vehicle.

mikaelacaron:
Closed with #32

PR #26, opened 2023-10-01
Fixed the bug, which was causing the vehicles, with no details added

ayushs-2k4:
Now, to add a vehicle, its name make, and model must have at least one character.

# What it Does
* Closes issue #24 
* It checks for the name, make, and model of a vehicle before adding it.

# How I Tested
For example:
* Run the application
* Tap the 'Settings' button
* Tap on 'Add Vehicle' Button
* Do not fill any details and tap 'Add'.
* Vehicle will not be added .
* Now tap on  'Add Vehicle' Button again.
* This time, fill some details.
* Tap on 'Add Button'
* Vehicle will get added.

# Notes
* Added check for name, make, and model strings, whether they are empty or not.

### DIFF
--- a/Basic-Car-Maintenance.xcodeproj/project.pbxproj
+++ b/Basic-Car-Maintenance.xcodeproj/project.pbxproj
@@ -53,7 +53,6 @@
 /* End PBXContainerItemProxy section */
 
 /* Begin PBXFileReference section */
-		FF098EFA2AB3424E003EC0FE /* Basic-Car-Maintenance.xcconfig */ = {isa = PBXFileReference; lastKnownFileType = text.xcconfig; path = "Basic-Car-Maintenance.xcconfig"; sourceTree = SOURCE_ROOT; };
 		FF09FC902AB6FF44006BE61A /* AuthenticationView.swift */ = {isa = PBXFileReference; lastKnownFileType = sourcecode.swift; path = AuthenticationView.swift; sourceTree = "<group>"; };
 		FF3DDF512AA4D28F009D91C4 /* DashboardViewModel.swift */ = {isa = PBXFileReference; lastKnownFileType = sourcecode.swift; path = DashboardViewModel.swift; sourceTree = "<group>"; };
 		FF5D13A32A86C2D600BC9BD6 /* Basic-Car-Maintenance.app */ = {isa = PBXFileReference; explicitFileType = wrapper.application; includeInIndex = 0; path = "Basic-Car-Maintenance.app"; sourceTree = BUILT_PRODUCTS_DIR; };
@@ -132,7 +131,6 @@
 		FF5D139A2A86C2D500BC9BD6 = {
 			isa = PBXGroup;
 			children = (
-				FF098EFA2AB3424E003EC0FE /* Basic-Car-Maintenance.xcconfig */,
 				FF5D13A52A86C2D600BC9BD6 /* Basic-Car-Maintenance */,
 				FF5D13B72A86C2D800BC9BD6 /* Basic-Car-MaintenanceTests */,
 				FF5D13C12A86C2D800BC9BD6 /* Basic-Car-MaintenanceUITests */,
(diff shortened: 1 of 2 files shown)


Basic-Car-Maintenance.xcodeproj/project.pbxproj
```diff
@@ -53,7 +53,6 @@
 /* End PBXContainerItemProxy section */
 
 /* Begin PBXFileReference section */
-		FF098EFA2AB3424E003EC0FE /* Basic-Car-Maintenance.xcconfig */ = {isa = PBXFileReference; lastKnownFileType = text.xcconfig; path = "Basic-Car-Maintenance.xcconfig"; sourceTree = SOURCE_ROOT; };
```
mikaelacaron, line 56: This shouldn't be removed, please follow the instructions again in the [README](https://github.com/mikaelacaron/Basic-Car-Maintenance/tree/main#getting-started) for properly setting up the project.

Basic-Car-Maintenance/Shared/Settings/Views/AddVehicleView.swift
```diff
@@ -42,7 +41,11 @@ struct AddVehicleView: View {
                 ToolbarItem {
                     Button {
                         let vehicle = Vehicle(name: name, make: make, model: model)
-                        addTapped(vehicle)
+                        
+                        if !name.isEmpty && !make.isEmpty && !model.isEmpty {
+                            addTapped(vehicle)
+                        }
+
```
mikaelacaron: A better way to do this is make this a computed property `!name.isEmpty && !make.isEmpty && !model.isEmpty` and disable the Add button until those parameters are met

Otherwise the user will be tapping Add and will be confused why nothing is happening

mikaelacaron (changes requested):
Thanks for contributing! Be sure you set the project up properly (you removed a file that should not have been removed, and I suggested a better way to optimize your code)

When you've addressed all the comments, tap the re-review button seen here:
Screenshot 2023-10-01 at 3 21 25 PM

Basic-Car-Maintenance/Shared/Settings/Views/AddVehicleView.swift
ayushs-2k4: Made the requested changes.

Basic-Car-Maintenance.xcodeproj/project.pbxproj
mikaelacaron: @ayushs-2k4 This file is still being changed, and it should not be changed at all.

See [this video](https://www.youtube.com/live/60ivxAOUrYU?si=NX8hVyb55QQUo-Gs&t=1667) at `27:47`

Basic-Car-Maintenance/Shared/Settings/Views/AddVehicleView.swift
```diff
@@ -43,16 +45,17 @@ struct AddVehicleView: View {
...
+                    .disabled(!isVehicleValid)
                 }
             }
         }
     }
 }
 
 #Preview {
-    AddVehicleView() { _ in }
+    AddVehicleView { _ in }
```
mikaelacaron, line 60: No need to change this

Basic-Car-Maintenance/Shared/Settings/Views/AddVehicleView.swift
```diff
@@ -43,16 +45,17 @@ struct AddVehicleView: View {
                     Button {
                         let vehicle = Vehicle(name: name, make: make, model: model)
                         addTapped(vehicle)
-                        dismiss()
+                        dismiss()
```
mikaelacaron, line 48: Also in the future be careful about adding extra whitespace

mikaelacaron (changes requested):
Follow the directions from the comment I made, the `pbxproj` file shouldn't be changed at all

Issue #12, opened 2023-09-29
IMPROVE - Error Handling When Deleting a Vehicle


mikaelacaron:
## Describe the feature you to improve
* In func `deleteVehicle` in the `SettingsViewModel` it doesn't handle if an error is thrown
* Add a do-catch block to show an error in an alert if an error is thrown

mishapark:
Hi, I'd like to work on this one

mishapark:
Please assign this one to me :)

mikaelacaron:
You got it!

mikaelacaron:
Closed with #29

PR #25, opened 2023-10-01
IMPROVE - Error Handling When Creating a Vehicle

samuelowino:
Issue #13

# What it Does
* Closes issue #13

```markdown
# Localization Update and Error Handling Change: Issue #XX

## What it Does
* Closes Issue #13
* Added new strings to the localizable file for better user-facing messages.
* Updated the ```addVehicle``` function to include proper error handling.

## Updated Localizable Strings
Here are the new strings that have been added to the Localizable.strings file:

```plaintext
"Failed to add vehicle" = "";
"Failed to add vehicle, internal error occurred. Unknown error occurred" = "";
"Ok, got it!" = "";
"Failed to add vehicle, internal error occurred\nDetails:%@" = "";
```

## Code Change Explanation
The `addVehicle` function now incorporates robust error handling. Here's a brief overview of the changes made:

```swift
func addVehicle(_ vehicle: Vehicle, then: @escaping(Result<String, Error>) -> ()) async {
    if let uid = authenticationViewModel.user?.uid {
        var vehicleToAdd = vehicle
        vehicleToAdd.userID = uid
        
        do {
            try Firestore
                .firestore()
                .collection("vehicles")
                .addDocument(from: vehicleToAdd)
            
            vehicles.append(vehicleToAdd)
            then(.success("Vehicle Added"))
        } catch {
            then(.failure(error))
        }
    }
}
```
----
In this updated code, the function now properly handles potential errors that may occur during the process of adding a vehicle to the Firestore database.

## How to Test
1. Run the application.
2. Trigger the `addVehicle` function with a valid `Vehicle` object.
3. Ensure that the vehicle is successfully added to the Firestore database.
4. Verify that the success message is returned through the completion handler.

## Notes
To force an error in ```AddVehicle``` use:

```swift
func addVehicle(_ vehicle: Vehicle, then: @escaping(Result<String, Error>) -> ()) async {
        
        if let uid = authenticationViewModel.user?.uid {
            var vehicleToAdd = vehicle
            vehicleToAdd.userID = uid
            
            do {
                try Firestore
                    .firestore()
                    .collection("vehicles")
                    .addDocument(from: vehicleToAdd)
                
                vehicles.append(vehicleToAdd)
                then(.failure(URLError(URLError.Code.badURL))) //throws badURL error
            } catch {
                then(.failure(error))
            }
            
        }
    }
```
## Screenshot
If applicable, include a screenshot or recording showcasing the updated localized strings and error handling in action.

PR #29, opened 2023-10-01
Added error handling for deleting vehicles

mishapark:
# What it Does
* Closes issue #12 
* Adds swipe actions for vehicles to delete them
* Adds error handling and an alert for deleting vehicles

# How I Tested
* Add a vehicle
* Swipe left and tap on Delete
* Add this line to `deleteVehicle` imitate an error: `then(.failure(URLError(URLError.Code.badURL)))`
* When deleting, the app shows an alert

# Notes
* Made the implementation similar to https://github.com/mikaelacaron/Basic-Car-Maintenance/pull/25 to have consistency in the code
* In the PR above `await` needs to be added before adding a vehicle to the Firebase, otherwise it displays a vehicle on the screen and when trying to delete it right after creation, the app shuts down

# Screenshot

### DIFF
--- a/Basic-Car-Maintenance/Shared/Settings/ViewModels/SettingsViewModel.swift
+++ b/Basic-Car-Maintenance/Shared/Settings/ViewModels/SettingsViewModel.swift
@@ -56,9 +56,9 @@ final class SettingsViewModel: ObservableObject {
         }
     }
     
-    func deleteVehicle(_ vehicle: Vehicle, then: @escaping(Result<String, Error>) -> Void) async {
+    func deleteVehicle(_ vehicle: Vehicle) async throws {
       guard let documentId = vehicle.id else {
-          fatalError("Event \(vehicle.name) has no document ID.")
+        fatalError("Event \(vehicle.name) has no document ID.")
       }
 
       do {
@@ -69,9 +69,8 @@ final class SettingsViewModel: ObservableObject {
           .delete()
 
         vehicles.removeAll { $0.id == vehicle.id }
-        then(.success("Vehicle deleted"))
       } catch {
-        then(.failure(error))
+        throw error
       }
     }
 }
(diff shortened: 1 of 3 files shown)


Basic-Car-Maintenance/Shared/Settings/ViewModels/SettingsViewModel.swift
```diff
@@ -56,14 +56,22 @@ final class SettingsViewModel: ObservableObject {
...
-    func deleteVehicle(_ vehicle: Vehicle) async {
-        guard let documentId = vehicle.id else {
-            fatalError("Event \(vehicle.name) has no document ID.")
-        }
-        try? await Firestore
-            .firestore()
-            .collection("vehicles")
-            .document(documentId)
-            .delete()
+    func deleteVehicle(_ vehicle: Vehicle, then: @escaping(Result<String, Error>) -> Void) async {
```
mikaelacaron: Can you do this using async await and throws, rather than using a closure?

mishapark:
Hi, @mikaelacaron, fixed as per your last comment. Thanks!

Basic-Car-Maintenance/Shared/Settings/Views/SettingsView.swift
```diff
@@ -65,6 +67,19 @@ struct SettingsView: View {
...
+                              Task {
+                                do {
+                                  try await viewModel.deleteVehicle(vehicle)
+                                } catch {
+                                  errorDetails = error
+                                  showDeleteVehicleError = true
+                                }
+                              }
+                            }
+                            .tint(.red)
```
mikaelacaron, line 81: Note that you don't need to set the `.tint` to red because you can instead give the Button `role: .destructive`

mikaelacaron (approved):
Great work!! Note that your changes all use `2` spaces, please use `4` for any future PRs because that's what this project uses.

Issue #28, opened 2023-10-01
FEATURE - Add description on dashboard screen


daljeetseera:
## Describe the feature you want the app to have that it currently doesn't
* Currently, the app opens with blank dashboard. I would like to add a description which helps user to take an action. 

## Describe alternatives you've considered
* A description with ```Add maintenance``` button which directly opens the add sheet.

## Additional context
* Screen must have a description with/without an action button.

daljeetseera:
Can I start on this one? @mikaelacaron

mikaelacaron:
I think a description would be best! Adding another button could be confusing

mikaelacaron:
Closed with #31

PR #31, opened 2023-10-01
Added description on dashboard screen and implemented delete event fr…

daljeetseera:
Feature #28 

# What it Does
* Closes issue #28 
* Implemented the description for Dashboard screen in case of no events.
* Delete event from events fro UI updates.

# How I Tested
* Tested the app on simulator with following steps
1) Run the app with no events. The description should display.
2) Add event and description should hide.
3) Delete event and description should display in case of 0 events.

# Notes
* I haven't added my team ID.

# Screenshot

https://github.com/mikaelacaron/Basic-Car-Maintenance/assets/54598657/0979b29c-6e98-428e-a522-b43ec8da3ab3

### DIFF
--- a/Basic-Car-Maintenance/Shared/Dashboard/ViewModels/DashboardViewModel.swift
+++ b/Basic-Car-Maintenance/Shared/Dashboard/ViewModels/DashboardViewModel.swift
@@ -64,5 +64,9 @@ class DashboardViewModel: ObservableObject {
             .collection("maintenance_events")
             .document(documentId)
             .delete()
+
+        if let index = events.firstIndex(where: {$0.id == documentId}) {
+            events.remove(at: index)
+        }
     }
 }
--- a/Basic-Car-Maintenance/Shared/Dashboard/Views/DashboardView.swift
+++ b/Basic-Car-Maintenance/Shared/Dashboard/Views/DashboardView.swift
@@ -41,6 +41,11 @@ struct DashboardView: View {
                 }
                 .listStyle(.inset)
             }
+            .overlay(Group {
+                if viewModel.events.isEmpty {
+                    Text("Add your first maintenance")
+                }
+            })
             .navigationTitle(Text("Dashboard"))
             .sheet(isPresented: $isShowingAddView) {
                 AddMaintenanceView() { event in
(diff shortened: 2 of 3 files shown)


Basic-Car-Maintenance/Shared/Dashboard/ViewModels/DashboardViewModel.swift
```diff
@@ -64,5 +64,9 @@ class DashboardViewModel: ObservableObject {
             .collection("maintenance_events")
             .document(documentId)
             .delete()
+
+        if let index = events.firstIndex(where: {$0.id == documentId}) {
+            events.remove(at: index)
+        }
```
mikaelacaron, line 70: This actually doesn't need to be fixed, because for 1, it's out of the scope of the issue you were working on (so be sure to stay on task), but 2nd because the button role is `.destructive` where this is used on the `DashboardView` I believe this happens automatically

I'm going to merge your PR, but remove this, because it's out of scope for this issue

If you think this is still an issue, file a bug to work on!

Basic-Car-Maintenance/Shared/Dashboard/Views/DashboardView.swift
```diff
@@ -41,6 +41,11 @@ struct DashboardView: View {
                 }
                 .listStyle(.inset)
             }
+            .overlay(Group {
+                if viewModel.events.isEmpty {
+                    Text("Add your first maintenance")
+                }
+            })
```
mikaelacaron, line 48: I'm going to merge this because this is a small change, but note that this doesn't need to be in a `Group` also be aware of your spacing, this is how this code should be spaced properly
```swift
.overlay {
    if viewModel.events.isEmpty {
        Text("Add your first maintenance")
    }
}
```

mikaelacaron (approved):
Approving this @daljeetseera but but the comments I have for next time!

Basic-Car-Maintenance/Shared/Dashboard/Views/DashboardView.swift
daljeetseera: Sure. I will keep that in mind.

Basic-Car-Maintenance/Shared/Dashboard/ViewModels/DashboardViewModel.swift
daljeetseera: I made this change because the ```StateObject viewmodel``` in ```DashboardView``` is not updating the UI until we update the events object.

Issue #35, opened 2023-10-02
IMPROVE - iOS 17 @Observable


Drag0ndust:
## Describe the feature you want to improve
As the project uses iOS17 as minimum target change all the @Published property wrapper to the new Observation Framework

Drag0ndust:
I want to do this task, if it is ok @mikaelacaron

mikaelacaron:
Closed with #36

PR #36, opened 2023-10-02
#35: New observation framework

Drag0ndust:
# What it Does
* Closes issue #35

Changed all view models to the new iOS17 Observation Framework

# Test
I testes all the features in the app and they are still working after the change


Basic-Car-Maintenance/Shared/BasicCarMaintenanceApp.swift
```diff
@@ -8,20 +8,14 @@
...
 @main
 struct BasicCarMaintenanceApp: App {
-    @UIApplicationDelegateAdaptor(AppDelegate.self) var delegate
+    @State private var authenticationViewModel: AuthenticationViewModel
     
-    @StateObject var authenticationViewModel = AuthenticationViewModel()
+    init() {
+        FirebaseApp.configure()
+        self._authenticationViewModel = .init(initialValue: AuthenticationViewModel())
+    }
```
mikaelacaron: We still need the `AppDelegate` this is the proper way from Firebase for using it in their SwiftUI lifecycle apps

Basic-Car-Maintenance/Shared/BasicCarMaintenanceApp.swift
```diff
@@ -8,20 +8,14 @@
...
-
 @main
 struct BasicCarMaintenanceApp: App {
-    @UIApplicationDelegateAdaptor(AppDelegate.self) var delegate
+    @State private var authenticationViewModel: AuthenticationViewModel
     
-    @StateObject var authenticationViewModel = AuthenticationViewModel()
+    init() {
+        FirebaseApp.configure()
+        self._authenticationViewModel = .init(initialValue: AuthenticationViewModel())
```
mikaelacaron: Is there any benefit to initializing it in the `init` rather than if it was written `@State private var authenticationViewModel = AuthenticationViewModel()`?

Basic-Car-Maintenance/Shared/Settings/ViewModels/AuthenticationViewModel.swift
```diff
@@ -22,21 +22,22 @@ enum AuthenticationFlow {
...
-    @Published var user: User?
+    var user: User?
     
-    @Published var flow: AuthenticationFlow = .signUp
+    var flow: AuthenticationFlow = .signUp
     
     private var authStateHandler: AuthStateDidChangeListenerHandle?
     private var currentNonce: String?
     
+    @MainActor
```
mikaelacaron, line 40: Why is the `@MainActor` here rather than keeping it at the class level? Same questions for all the functions in this class

Basic-Car-Maintenance/Shared/Settings/Views/AuthenticationView.swift
```diff
@@ -11,10 +11,10 @@ import AuthenticationServices
 struct AuthenticationView: View {
     @Environment(\.colorScheme) var colorScheme
     
-    @ObservedObject var viewModel: AuthenticationViewModel
+    @State var viewModel: AuthenticationViewModel
```
mikaelacaron: This this still `@State` or is it now `@Bindable`? because we were passing a reference to the class into this view.
Also with this, you're creating a new instance of the viewModel it looks like in the `init` rather than using the one instance defined at the top level of the app

Basic-Car-Maintenance/Shared/Settings/Views/SettingsView.swift
```diff
@@ -8,14 +8,11 @@
...
-    
-    @StateObject private var viewModel: SettingsViewModel
+    @State private var viewModel: SettingsViewModel
     @State private var isShowingAddVehicle = false
-    @ObservedObject var authenticationViewModel: AuthenticationViewModel
     
     init(authenticationViewModel: AuthenticationViewModel) {
-        self._viewModel = StateObject(wrappedValue: SettingsViewModel(authenticationViewModel: authenticationViewModel)) // swiftlint:disable:this line_length
-        self.authenticationViewModel = authenticationViewModel
+        viewModel = SettingsViewModel(authenticationViewModel: authenticationViewModel)
```
mikaelacaron, line 17: Same question here as before about creating this here vs in the View

mikaelacaron (changes requested):
Great work @Drag0ndust ! I have a few questions about how some of the new macros work thanks!

Basic-Car-Maintenance/Shared/BasicCarMaintenanceApp.swift
Drag0ndust: The new SwiftUI lifecycle which uses the init method. It is basically the same as the didFinishLaunching in the old lifecycle. But I can revert it if you want

Basic-Car-Maintenance/Shared/BasicCarMaintenanceApp.swift
Drag0ndust: The 'Firebase.init()' has to be called first, that's why I changed it here

Basic-Car-Maintenance/Shared/BasicCarMaintenanceApp.swift
mikaelacaron: Can you link documentation that says this is the same thing?
Because previously it wasn't, so I wanted to double check, also because the AppDelegate has many other things it does, I'm confused how all of that now is just in `init`

Drag0ndust: Here is the part in the [Apple documentation](https://developer.apple.com/documentation/swiftui/uiapplicationdelegateadaptor): Important hint in the middle

Here is an explanation from Peter Friese [how you can initialise Firebase](https://stackoverflow.com/a/62633158)

But if you want I can switch it back to the "old" way with the AppDelegate.

Basic-Car-Maintenance/Shared/Settings/ViewModels/AuthenticationViewModel.swift
Drag0ndust: Only some methods need to use the Mainthread/MainActor, that's why I just added it to those parts

Basic-Car-Maintenance/Shared/Settings/Views/AuthenticationView.swift
Drag0ndust: You are right, see [Apple Documentation](https://developer.apple.com/documentation/swiftui/migrating-from-the-observable-object-protocol-to-the-observable-macro)

Will push the change in a few minutes

Drag0ndust:
Answered your comments, please have a look @mikaelacaron

Basic-Car-Maintenance/Shared/BasicCarMaintenanceApp.swift
mikaelacaron: That's what I wanted to see thanks!

mikaelacaron (approved):
Great work! Thanks for all the links and implementing the suggestions

mikaelacaron:
Hey @Drag0ndust  I think I may have merged this wrong haha 😅
Would you be able to make a new branch and move your changes onto that (with the latest updates from `dev`)
haha sorry about that

Drag0ndust:
@mikaelacaron now it should be ok, please recheck it and merge it then 😀

Basic-Car-Maintenance/Shared/Dashboard/Views/DashboardView.swift
```diff
@@ -9,10 +9,11 @@ import SwiftUI
 
 struct DashboardView: View {
     
-    @StateObject private var viewModel: DashboardViewModel
+    @State private var isShowingAddView = false
+    @Bindable private var viewModel: DashboardViewModel
```
mikaelacaron, line 12: I think the `DashboardViewModel` needs to be `@State`

Basic-Car-Maintenance/Shared/Settings/ViewModels/SettingsViewModel.swift
```diff
@@ -51,7 +50,7 @@ final class SettingsViewModel: ObservableObject {
             var vehicleToAdd = vehicle
             vehicleToAdd.userID = uid
             
-            try? Firestore
+            _ = try? Firestore
```
mikaelacaron:
```suggestion
            try? Firestore
```

Basic-Car-Maintenance/Shared/Settings/Views/AuthenticationView.swift
```diff
@@ -11,7 +11,7 @@ import SwiftUI
 struct AuthenticationView: View {
     @Environment(\.colorScheme) var colorScheme
     
-    @ObservedObject var viewModel: AuthenticationViewModel
+    var viewModel: AuthenticationViewModel
```
mikaelacaron, line 14: I think this needs to be `@Bindable`

Basic-Car-Maintenance/Shared/Settings/Views/ContributorsListView.swift
```diff
@@ -9,7 +9,11 @@ import SwiftUI
 
 struct ContributorsListView: View {
     
-    @ObservedObject var viewModel: SettingsViewModel
+    var viewModel: SettingsViewModel
```
mikaelacaron, line 12: I think this needs to be `@Bindable`

Basic-Car-Maintenance/Shared/Settings/Views/SettingsView.swift
```diff
@@ -8,16 +8,13 @@
 import SwiftUI
 
 struct SettingsView: View {
-    
-    @StateObject private var viewModel: SettingsViewModel
+    private var viewModel: SettingsViewModel
```
mikaelacaron: Does this need to be `@Bindable`

mikaelacaron (changes requested):
We discussed on livestream some code I'm not sure if it should be `@Bindable` or not
@elmoritz can you drop the docs you found?

elmoritz:
According to the [documentation](https://developer.apple.com/documentation/swiftui/migrating-from-the-observable-object-protocol-to-the-observable-macro) you only need the `@Bindable` if you want to access a property on that object via `$`.

look at the last example.😎

Basic-Car-Maintenance/Shared/Settings/Views/AuthenticationView.swift
elmoritz: here we do not use `$viewModel.xxxxx`
So according to the docs, this should not be `@Bindable`

Basic-Car-Maintenance/Shared/Settings/Views/ContributorsListView.swift
elmoritz: here we do not use `$viewModel.xxxxx`
So according to the docs, this should not be `@Bindable`

Basic-Car-Maintenance/Shared/Settings/Views/SettingsView.swift
elmoritz: Same as above:
here we do not use `$viewModel.xxxxx`
So according to the docs, this should not be `@Bindable`

Basic-Car-Maintenance/Shared/Settings/ViewModels/SettingsViewModel.swift
Drag0ndust: This would add a warning. What is the point of this change?

Basic-Car-Maintenance/Shared/Dashboard/Views/DashboardView.swift
Drag0ndust: No, here we need `@Bindable` (see the link @elmoritz provided)

Drag0ndust:
I applied all suggested changes and checked all other questions with the documentation link Moritz provided and now it should be fine.

mikaelacaron:
@Drag0ndust  pull changes again, I just pushed new changes for fixing all the signing stuff
Thank so much for all your work on this! (I'll probably need to pull changes again, cause there are some new PRs) that I would want to merge before merging this one with all the new observation stuff

Drag0ndust:
Ok, then notify me when you merged all those things, otherwise I have to merge several times. So I can save time on my side.

Basic-Car-Maintenance/Shared/Dashboard/Views/DashboardView.swift
```diff
@@ -9,12 +9,12 @@ import SwiftUI
 
 struct DashboardView: View {
     
-    @StateObject private var viewModel: DashboardViewModel
-    @State private var isShowingEditView = false
+    @Bindable private var viewModel: DashboardViewModel
+    @State private var isShowingAddView = false
```
mikaelacaron: I think this was merged wrong, there should be `isShowingEditView`, for editing (but I think an extra property also got left, so I'm going to fix it and merge this

Basic-Car-Maintenance/Shared/Settings/Views/ContributorsListView.swift
```diff
@@ -9,7 +9,11 @@ import SwiftUI
...
+
+    init(viewModel: SettingsViewModel) {
```
mikaelacaron: Removing this `init` cause it's not needed, structs create their own `init`s and there's nothing special about this one

mikaelacaron (approved):
Great work!! Thanks for fixing multiple merge conflicts
I'm going to push to your branch and then merge this!

Issue #7, opened 2023-09-29
FEATURE - Localization (Multiple assignees allowed!)


mikaelacaron:
## Describe the feature you want the app to have that it currently doesn't
* Using the new `Localizable.xcstrings` (string catalogs) translate the app into any language that is missing
* Multiple assignees are allowed for this issue!
* Comment what language you are translating to, so that two people aren't doing the same language.

## Describe alternatives you've considered
* This will be an ongoing issue that will continually change as the app is developed. If you create a new language and later more keys are added, you are welcome to make a new PR with updates to the translations.

## Additional context
* See [this video](https://developer.apple.com/wwdc23/10155) to learn how the new string catalogs work

appfrosch:
I'd like to work on German localization on an ongoing basis. German is my native language.

harryheres:
I could do Czech localization as my native language is Czech :)

sergpryimachuk:
I can help with Ukrainian, it's my native tongue.

mikaelacaron:
@appfrosch @HarryHeres @sergpryimachuk Go for it!! There's many changes upcoming today, so be aware some merge conflicts may arise soon!
And throughout the project you're free to check back in and make a PR for any translations that may be missing as the app continues to be developed

Angel39706:
Hello Mikaela, May I help with Spanish ?

narendrajagne:
Yes ,  I will help u with Hindi (INDIA)

arirays:
Hi, I'd like to work on Russian localization. Thanks!

dt-thomas-durand:
Hi @mikaelacaron.
I can do French if you want :)

mikaelacaron:
@dt-thomas-durand Yes please!

josephnoir:
I'd like to do Korean.

ilyapaddubny:
I can do the Belarusian (be) localization

AsifSyeed:
I can do Bengali

horacio3m:
id love to work on Brazilian Portuguese localization.  @mikaelacaron

isiko35:
I can work on Turkish localization @mikaelacaron

maartinj:
Hi @mikaelacaron, I can help with Polish language.

mikaelacaron:
@maartinj go for it! Thank you!

lonnekemennen:
I can translate to Dutch! 🇳🇱

nssina:
[...]

PR #37, opened 2023-10-02
7 localize german

appfrosch:
# What it Does
* adding German translations (see issue #7)

# How I Tested
* Ran the application
* Set the Sim device to German
* Checked all the currently available screens

# Notes
* it will be an ongoing effort to check for new strings to translate during development
* I will check in regularily

# Screencast
https://github.com/mikaelacaron/Basic-Car-Maintenance/assets/91844195/625edc0b-b348-4bbf-9158-0c9171bb8796

### DIFF
--- a/Basic-Car-Maintenance.xcodeproj/project.pbxproj
+++ b/Basic-Car-Maintenance.xcodeproj/project.pbxproj
@@ -358,6 +358,7 @@
 			knownRegions = (
 				en,
 				Base,
+				de,
 			);
 			mainGroup = FF5D139A2A86C2D500BC9BD6;
 			packageReferences = (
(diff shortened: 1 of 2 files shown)


Basic-Car-Maintenance.xcodeproj/project.pbxproj
```diff
@@ -623,7 +624,7 @@
 				MACOSX_DEPLOYMENT_TARGET = 13.4;
 				MARKETING_VERSION = 1.0;
 				OTHER_LDFLAGS = "-ObjC";
-				PRODUCT_BUNDLE_IDENTIFIER = "com.icyappstudio.Basic-Car-Maintenance";
+				PRODUCT_BUNDLE_IDENTIFIER = "ch.appfros.Basic-Car-Maintenance";
```
mikaelacaron: Don't change the bundle ID please! Go back and fix all the instances of this

mikaelacaron (changes requested):
Please fix the bundle ID change and then click the review button

Re-review button

Basic-Car-Maintenance.xcodeproj/project.pbxproj
appfrosch: Sorry, that's quite obviously not a good idea... As per our small conversation on Mastadon, I changed it cause it gave me an error for macOS–but committing it in the PR doesn't make sense anyway…

mikaelacaron (approved):
Great work!

PR #38, opened 2023-10-02
Added Czech localization

harryheres:
# What it Does
* Adds a new localization option for the Czech language

# How I Tested
* Went through the whole app and checked every string present to the user.

# Notes
* None

# Screenshots
image

image

### DIFF
--- a/Basic-Car-Maintenance/Shared/Localizable.xcstrings
+++ b/Basic-Car-Maintenance/Shared/Localizable.xcstrings
@@ -2,91 +2,293 @@
   "sourceLanguage" : "en",
   "strings" : {
     "" : {
-
+      "localizations" : {
+        "cs" : {
+          "stringUnit" : {
+            "state" : "translated",
+            "value" : ""
+          }
+        }
+      }
     },
     "%@" : {
-
+      "localizations" : {
+        "cs" : {
+          "stringUnit" : {
+            "state" : "translated",
+            "value" : "%@"
+          }
+        }
+      }
     },
     "Add" : {
-
+      "localizations" : {
+        "cs" : {
+          "stringUnit" : {
+            "state" : "translated",
+            "value" : "Přidat"
+          }
+        }
+      }
     },
     "Add Maintenance" : {
-
+      "localizations" : {
+        "cs" : {
+          "stringUnit" : {
+            "state" : "translated",
+            "value" : "Přidat servisní záznam"
+          }
+        }
+      }
     },
     "Add Vehicle" : {
-
+      "localizations" : {
+        "cs" : {
+          "stringUnit" : {
+            "state" : "translated",
+            "value" : "Přidat vozidlo"
+          }
+        }
+      }
     },
     "Additional Notes" : {
-
+      "localizations" : {
+        "cs" : {
+          "stringUnit" : {
+            "state" : "translated",
+            "value" : "Další poznámky"
+          }
+        }
+      }
     },
     "Dashboard" : {
-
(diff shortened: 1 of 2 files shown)


mikaelacaron (changes requested):
@HarryHeres can you solve these merge conflicts please? unfortunately it looks like you may have to fix each translation one by one, in some kind of editor, because I'm guessing this wont't build in Xcode, cause it can't parse the Localizable.xcstrings file

harryheres:
@mikaelacaron Conflicts fixed :). This might happen also in the future, since there are many of us working on the translations.

mikaelacaron:
Yeah it's definitely going to happen again, did you solve it by using an editor outside of Xcode?

harryheres:
> Yeah it's definitely going to happen again, did you solve it by using an editor outside of Xcode?

I do all of my other work in Neovim, so fired that up and manually fixed all. I suppose as more languages will get added and get translated, the amount of merge conflicts will eventually lower. But at the moment, we will need to resolve these manually, becase the Localization source file is just a JSON in the end, so after resolving it, I recommend using a JSON analyzer to check the structure. If it is fine, XCode will accept it and build it out.

Basic-Car-Maintenance/Shared/Localizable.xcstrings
```diff
@@ -43,29 +67,48 @@
     },
     "Add Vehicle" : {
       "localizations" : {
-        "de" : {
+        "cs" : {
           "stringUnit" : {
             "state" : "translated",
-            "value" : "Fahrzeug hinzufügen"
+            "value" : "Přidat vozidlo"
```
mikaelacaron: Is this overwriting the German translation?

Basic-Car-Maintenance/Shared/Localizable.xcstrings
```diff
@@ -43,29 +67,48 @@
...
+            "state" : "translated",
+            "value" : "Další poznámky"
+          }
+        },
         "de" : {
           "stringUnit" : {
             "state" : "translated",
-            "value" : "Weitere Notizen"
+            "value" : "Fahrzeug hinzufügen"
           }
```
mikaelacaron: Is this overwriting the German translation?

mikaelacaron (changes requested):
Just two questions!
Yeah this makes sense that there will be merge conflicts, it depends on when someone forks the project and makes the translation

Basic-Car-Maintenance/Shared/Localizable.xcstrings
harrythewise: Oh, it seems like I must’ve missed a line! Will fix that tomorrow. Sorry!

Basic-Car-Maintenance/Shared/Localizable.xcstrings
harrythewise: Oh, it seems like I must’ve missed a line! Will fix that tomorrow. Sorry!

Basic-Car-Maintenance/Shared/Localizable.xcstrings
harrythewise: Fixed.

Basic-Car-Maintenance/Shared/Localizable.xcstrings
harrythewise: Fixed.

mikaelacaron (approved):
Great work! Thanks for solving the merge conflicts!

PR #39, opened 2023-10-02
Fix issue - #10 : Error Handling When Deleting an Event

### LINKED ISSUE #10: IMPROVE - Error Handling When Deleting an Event
* The `deleteEvent` function in the `DashboardViewModel` doesn't handle errors and uses `try?`
* Use a do-catch block for this instead, and show an alert with the error's localized description, if something fails

Dishant10:
# What it Does
* Closes #10
* Handled error in func deleteEvent, added do catch block and added two variables which are responsible for showing an alert and error message with that alert. 

# How I Tested
* Tested by setting the showAlert variable to true while deleting an event and the alert was showing as soon as the an event was deleted from the list.

# Notes

# Screenshot
* This is not a feature but a simple case of error handling hence I do not know what to screenshot and add here. Let me if there's a specific test or use case I'll definitely add that.


Basic-Car-Maintenance/Shared/Dashboard/ViewModels/DashboardViewModel.swift
```diff
@@ -15,6 +15,8 @@ class DashboardViewModel: ObservableObject {
     let authenticationViewModel: AuthenticationViewModel
     
     @Published var events = [MaintenanceEvent]()
+    @Published var showErrorAlert = false
+    @Published var errorDeleting : Error?
```
mikaelacaron: Can you rename this `errorMessage`

Basic-Car-Maintenance/Shared/Dashboard/ViewModels/DashboardViewModel.swift
```diff
@@ -59,10 +61,17 @@ class DashboardViewModel: ObservableObject {
...
+            try await Firestore
+                .firestore()
+                .collection("maintenance_events")
+                .document(documentId)
+                .delete()
+        }
+        catch{
+            showErrorAlert.toggle()
+            errorDeleting = error
+            //print("Error : \(error.localizedDescription)")
```
mikaelacaron: Add a space between the curly brace and the words `do` and `catch`, and put the would catch on the line with the brace `}`
Also can you use the error's `localizedDescription`?

The statement would look like this
```swift
do {
} catch {
}
```

Basic-Car-Maintenance/Shared/Dashboard/Views/DashboardView.swift
```diff
@@ -50,6 +50,13 @@ struct DashboardView: View {
                     
                 }
             }
+            .alert("Failed To Delete Event", isPresented: $viewModel.showErrorAlert, actions: {
+                Button("OK") {
+                    viewModel.showErrorAlert = false
+                }
+            }, message: {
+                Text(viewModel.errorDeleting?.localizedDescription ?? "Error while deleting event").padding()
```
mikaelacaron: As noted above just pass the localized description, and handle the optional in the view model rather than here in the view
(This would mean the `errorDeleting` would actually end up being a `String` rather than `Error`)

mikaelacaron (changes requested):
Great work! Just a few changes, thanks!

When you've made the changes, click the "re-request" review button
Re-review button

Basic-Car-Maintenance/Shared/Dashboard/Views/DashboardView.swift
Dishant10: Got it! Pushed the code with the suggested changes. Let me know if the code needs more changes.

Basic-Car-Maintenance/Shared/Dashboard/ViewModels/DashboardViewModel.swift
```diff
@@ -52,17 +54,23 @@ class DashboardViewModel: ObservableObject {
...
+            try await Firestore
+                .firestore()
+                .collection("maintenance_events")
+                .document(documentId)
+                .delete()
+            errorMessage = ""
+        } catch {
+            showErrorAlert.toggle()
+            errorMessage = error.localizedDescription
+            //print("Error : \(error.localizedDescription)")
```
mikaelacaron, line 81: Delete this print next time, because it's commented out

mikaelacaron (approved):
Great work!

Issue #42, opened 2023-10-02
FEATURE - Delete a Vehicle


ayushs-2k4:
## Describe the feature you want the app to have that it currently doesn't
* Vehicle Deletion Feature

## Describe alternatives you've considered
* Currently, we cannot delete a vehicle once created. There should be a swipe to delete the vehicle feature in Settings View.

## Additional context
*N/A

ayushs-2k4:
I will be working on feature, once I get confirmation from @mikaelacaron that a vehicle can be deleted from Vehicles list or not able to delete a vehicle is an intended feature.

mikaelacaron:
Yes! This is an intended feature, go ahead and start it

mikaelacaron:
This functionality already exists

PR #45, opened 2023-10-02
Feature - delete a vehicle

ayushs-2k4:
# What it Does
* Closes issue #42 
* This change add Vehicle deletion capability through swiping.

# How I Tested
* Add a new vehicle, if there is no existing Vehicle.
* Swipe from right to left.
* Click on "Delete" Button.
* Vehicle will get deleted permanently.

For example:
* Run the application
* Tap the "Settings" Button.
* Tap "Add Vehicle" Button.
* Fill in the details of the new vehicle.
* Tap "Add".
* Vehicle will get added.
* Now to delete the vehicle, swipe from right to left, and click the "Delete" Button.
* Vehicle will get deleted.
* Restart the application to confirm that deletion has persisted.

# Notes
* N/A

# Screenshot
Simulator Screenshot - iPhone 15 Pro - 2023-10-03 at 01 52 09

### DIFF
--- a/Basic-Car-Maintenance/Shared/Settings/Views/SettingsView.swift
+++ b/Basic-Car-Maintenance/Shared/Settings/Views/SettingsView.swift
@@ -66,6 +66,15 @@ struct SettingsView: View {
                             Text(vehicle.model)
                         }
                     }
+                    .onDelete(perform: { indexSet in
+                        for index in indexSet {
+                            let vehicle = viewModel.vehicles[index]
+                            
+                            Task {
+                                await viewModel.deleteVehicle(vehicle)
+                            }
+                        }
+                    })
                     
                     Button {
                         isShowingAddVehicle.toggle()
--- a/Basic-Car-Maintenance/Shared/Settings/Views/SettingsView.swift
+++ b/Basic-Car-Maintenance/Shared/Settings/Views/SettingsView.swift
@@ -8,7 +8,6 @@
 import SwiftUI
 
 struct SettingsView: View {
-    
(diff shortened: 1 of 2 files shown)


Basic-Car-Maintenance/Shared/Settings/Views/SettingsView.swift
```diff
@@ -66,6 +66,15 @@ struct SettingsView: View {
...
                     }
+                    .onDelete(perform: { indexSet in
+                        for index in indexSet {
+                            let vehicle = viewModel.vehicles[index]
+                            
+                            Task {
+                                await viewModel.deleteVehicle(vehicle)
+                            }
+                        }
+                    })
```
mikaelacaron: Can you use instead specifically use this modifier `.swipeActions(edge: .trailing, allowsFullSwipe: true)`

So that it will be implemented the same way as the `Dashboard`

mikaelacaron (changes requested):
Thanks for the contribution! While this solution works, can you implement it a different way, so that it's implemented the same way as the `DashboardView`

Basic-Car-Maintenance/Shared/Settings/Views/SettingsView.swift
ayushs-2k4: Ok, let me change it

ayushs-2k4: I have changed onDelete to swipeActions. Kindly review it.

mikaelacaron:
Closing this PR because this functionality exists

PR #46, opened 2023-10-02
Fix Issue #11 - IMPROVE - Error Handling When Creating an Event

### LINKED ISSUE #11: IMPROVE - Error Handling When Creating an Event
* `try? await viewModel.addEvent(event)` is called in the `DashboardView` and doesn't handle if the function throws an error
* Handle this by wrapping it in a do-catch block, and showing an alert if there's an error
* Make sure `viewModel.events.append(event)` is *not* called when the error occurs

achi113s:
These changes add error handling for the addEvent(_ maintenanceEvent: MaintenanceEvent) method in DashboardViewModel. They also modify the sheet for the AddMaintenanceView in DashboardView to present the alert on dismissal if an error occurred (see notes).

# What it Does
* Closes #11
* Catch an error thrown from the following function call and alerts the user:
`
try Firestore.firestore().collection("maintenance_events").addDocument(from: eventToAdd)
`

# How I Tested
* Throw a sample error from the do block; I used a BadURL error.
For example:
* Run the application
* Tap the (+) button
* Tap the Add button from the AddMaintenanceEvent screen
Screenshot 2023-10-02 at 4 52 38 PM

# Notes
Modifying the .sheet() modifier in DashboardView is required to react to an error thrown from the AddMaintenanceEvent sheet because the DashboardView cannot present the sheet and the alert at the same time: 

`Attempt to present <SwiftUI.PlatformAlertController: 0x138980000> on <_TtGC7SwiftUI19UIHostingControllerGVS_15ModifiedContentVS_7AnyViewVS_12RootModifier__: 0x11f013c00> (from <_TtGC7SwiftUI32NavigationStackHostingControllerVS_7AnyView_: 0x138843c00>) which is already presenting <_TtGC7SwiftUI29PresentationHostingControllerVS_7AnyView_: 0x11f06c400>.
`

To get around this, I used the onDismissal attribute in .sheet() to toggle showing the alert if an error occurred.

I saw that @Dishant10 was working on adding error handling to deleteEvent(_ event: MaintenanceEvent) in this same file, but he shouldn't run into the same error because the error from deleteEvent is thrown when there is no sheet being presented. I tried to use similar variable names as him to reduce merging conflicts.

# Screenshot
Screenshot 2023-10-02 at 4 52 38 PM

### DIFF
--- a/Basic-Car-Maintenance/Shared/Dashboard/Views/DashboardView.swift
+++ b/Basic-Car-Maintenance/Shared/Dashboard/Views/DashboardView.swift
@@ -48,15 +48,14 @@ struct DashboardView: View {
             }
             .animation(.linear, value: viewModel.sortOption)
             .navigationTitle(Text("Dashboard"))
-            .sheet(isPresented: $isShowingAddView) {
+            .navigationDestination(isPresented: $isShowingAddView) {
                 AddMaintenanceView() { event in
                     Task {
                         await viewModel.addEvent(event)
                     }
-                    
                 }
             }
-            .alert("Failed To Delete Event", isPresented: $viewModel.showErrorAlert) {
+            .alert("An Error Occurred", isPresented: $viewModel.showErrorAlert) {
                 Button("OK") {
                     viewModel.showErrorAlert = false
                 }
--- a/Basic-Car-Maintenance/Shared/Dashboard/Views/DashboardView.swift
+++ b/Basic-Car-Maintenance/Shared/Dashboard/Views/DashboardView.swift
@@ -9,7 +9,6 @@ import SwiftUI
 
 struct DashboardView: View {
     
-    @State private var isShowingAddView = false
     @StateObject private var viewModel: DashboardViewModel
     
     init(authenticationViewModel: AuthenticationViewModel) {
@@ -48,7 +47,7 @@ struct DashboardView: View {
             }
             .animation(.linear, value: viewModel.sortOption)
             .navigationTitle(Text("Dashboard"))
-            .navigationDestination(isPresented: $isShowingAddView) {
+            .navigationDestination(isPresented: $viewModel.showAddView) {
                 AddMaintenanceView() { event in
                     Task {
                         await viewModel.addEvent(event)
@@ -65,7 +64,7 @@ struct DashboardView: View {
             .toolbar {
                 ToolbarItemGroup(placement: .primaryAction) {
                     Button {
-                        isShowingAddView.toggle()
+                        viewModel.showAddView.toggle()
                     } label: {
                         Image(systemName: "plus")
                     }
--- a/Basic-Car-Maintenance.xcodeproj/project.pbxproj
+++ b/Basic-Car-Maintenance.xcodeproj/project.pbxproj
@@ -443,7 +443,6 @@
 			knownRegions = (
 				en,
 				Base,
-				cs,
 				de,
 			);
 			mainGroup = FF5D139A2A86C2D500BC9BD6;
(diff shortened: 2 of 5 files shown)


Basic-Car-Maintenance/Shared/Dashboard/Views/DashboardView.swift
```diff
@@ -47,14 +47,24 @@ struct DashboardView: View {
                 }
             }
             .navigationTitle(Text("Dashboard"))
-            .sheet(isPresented: $isShowingAddView) {
+            .sheet(isPresented: $isShowingAddView, onDismiss: {
+                if viewModel.errorOccurred {
+                    viewModel.isShowingErrorAlert = true
+                }
+            }, content: {
```
mikaelacaron: I understand why, but I don't really like this as a solution, because now you have to remember that `errorOccurred` is essentially the same as `isShowingErrorAlert` and that will be forgotten

Can you try this [solution](https://stackoverflow.com/a/75795992/17309983)?
Or if that doesn't work, I'm fine with instead making the `AddMaintenanceView` be pushed onto the `DashboardView` so that you can then present an alert

Thanks for your work on this!

Also note I prefer multiple trailing closure syntax, so the alert modifier would look like this instead:

```swift
.alert("An Error Occurred", isPresented: $viewModel.isShowingErrorAlert) {
    Button("OK", role: .cancel) {
        viewModel.errorOccurred = false
    }
} message: {
    Text("\(viewModel.errorMessage)")
}
```

mikaelacaron (changes requested):
After you resolve the comment, click the re-review button thanks!

Re-review button

Basic-Car-Maintenance/Shared/Dashboard/Views/DashboardView.swift
achi113s: Thanks for suggesting that solution! Unfortunately it only partially worked. Here's what would happen when attaching .alert() to the AddMaintenanceView inside of the .sheet() :

- Go to AddMaintenanceView from Dashboard.
- Press "Add", test error URLError is thrown.
- Sheet dismisses but the alert is not shown.
- Go back to the AddMaintenanceView.
- Alert is now shown because isShowingErrorAlert was set to true when the error was thrown before.

Screenshot 2023-10-02 at 4 52 38 PM


So I went ahead with pushing AddMaintenanceView to the NavigationStack instead. Using .navigationDestination(isPresented: destination:) allows me to change less of the code, but the documentation says we should prefer binding NavStack to a Path. That would be more scalable, but we are only pushing a single view so maybe it's not necessary. Let me know if you want that instead!

-- Giorgio

achi113s:
Hmmm looks like some PRs got merged while I was working on this and now mine produces merge conflicts lol. How should I fix this? Delete my branch and redo my changes?

achi113s:
Alright, I fixed the merge conflicts. :) 

Literally accidentally closed this PR when I updated my fork to have all the new merged PRs lol. can you tell i'm new to this. :')

mikaelacaron:
> Alright, I fixed the merge conflicts. :)
> 
> Literally accidentally closed this PR when I updated my fork to have all the new merged PRs lol. can you tell i'm new to this. :')

You did great!! There's tons of changes right now with so many people

mikaelacaron (changes requested):
Good work! There's one small bug in this.
In the `AddMaintenanceView` the `dismiss()` is called right after the `addTapped` whether or not the error occurred.

This makes it so that the `AddMaintenanceView` dismisses and the alert is still shown, which is kinda odd, it should stay on the `AddMaintenanceView` and the alert is shown

achi113s:
> Good work! There's one small bug in this. In the `AddMaintenanceView` the `dismiss()` is called right after the `addTapped` whether or not the error occurred.
> 
> This makes it so that the `AddMaintenanceView` dismisses and the alert is still shown, which is kinda odd, it should stay on the `AddMaintenanceView` and the alert is shown

Should tapping on the OK button on the error alert dismiss the AddMaintenanceView or should we stay on AddMaintenanceView?

mikaelacaron:
> Should tapping on the OK button on the error alert dismiss the AddMaintenanceView or should we stay on AddMaintenanceView?

It should stay on the `AddMaintenanceView` and only dismiss the alert

Issue #52, opened 2023-10-03
Add GitHub Actions CI for SwiftLint


rizwankce:
## Describe the feature you want the app to have that it currently doesn't
* Since we get lot of PR and some time its easy to break the expected behaviour. To avoid it we can easily add CI (provided by GitHub Actions for free for OSS projects). CI can run tests on each PR and make sure all tests passes before we merge the new change in.

## Describe alternatives you've considered
* Alternatives would be adding via Circle CI or Bitraise but its needs extra steps and logins. Since GitHub Actions are free even for Mac OS machines its should be fine.

## Additional context
* There are lots of possibilities on this one. We can run Unit tests, UI test and even linting with swift lint and add the errors directly into the Github PR UI.

I have made the same config file and did a test run on my [repo](https://github.com/rizwankce/Basic-Car-Maintenance/pulls). Please take a look and if it aligns with your needs - I can raise a PR to this repo. Thanks for doing this awesome work.

rizwankce:
I would like to work on this. If you are okay with it @mikaelacaron

am2089:
If this has not been taken. I would like to work on this @mikaelacaron. Also, will you have more issues to work on?

mikaelacaron:
@rizwankce Thank you! I was thinking on how this could work. Can you make a PR with your CI script for running SwiftLint with GitHub Actions? (putting it in the proper place in the `.github` folder)
Run it when there's a PR made to `dev` or `main` branches please

@am2089 Yes more issues to come! So many people are concurrently working right now I'm debating which ones to add without creating too many merge conflicts.
You can set the repo to "watch" it where you can customize which notifications to get when I add new issues (I think).
But I'll create a new issue, for setting up GitHub actions for running unit tests

rizwankce:
Sure thing @mikaelacaron 

I have already made some changes on my fork. I'll raise a PR soon for review.

mikaelacaron:
Closed with #57

PR #57, opened 2023-10-03
Create swiftlint.yml

rizwankce:
# What it Does
* Closes issue #52 
* Adds new directory for GitHub actions
* Adds new config `swiftlint.yml` for Github actions

# How I Tested
* Tested locally at my machine and as well as GitHub Actions on my repo [here](https://github.com/rizwankce/Basic-Car-Maintenance/actions/runs/6390586029/job/17344138176)

### DIFF
--- a/.github/workflows/swiftlint.yml
+++ b/.github/workflows/swiftlint.yml
@@ -1,0 +1,16 @@
+name: SwiftLint
+on:
+  workflow_dispatch: {}
+  pull_request:
+    paths:
+      - '.github/workflows/swiftlint.yml'
+      - '.swiftlint.yml'
+      - '**/*.swift'
+jobs:
+  SwiftLint:
+    runs-on: macos-13
+    steps:
+      - name: Checkout
+        uses: actions/checkout@v4
+      - name: SwiftLint
+        run: swiftlint --quiet --reporter github-actions-logging


.github/workflows/swiftlint.yml
```diff
@@ -0,0 +1,16 @@
+name: SwiftLint
+on:
+  workflow_dispatch: {}
```
rizwankce, line 3: This is for arbitrary run this action via Github Web UI. So you can trigger linting for any branch you want from a single click.

.github/workflows/swiftlint.yml
```diff
@@ -0,0 +1,16 @@
...
+  pull_request:
+    paths:
+      - '.github/workflows/swiftlint.yml'
+      - '.swiftlint.yml'
+      - '**/*.swift'
```
rizwankce, line 8: This is general way of linting swift projects. It will run on any PR where the mentioned path are changed.

- If Ci config changed
- if lint YML file changed 
- If any swift file changed inside the project

.github/workflows/swiftlint.yml
```diff
@@ -0,0 +1,16 @@
...
+jobs:
+  SwiftLint:
+    runs-on: macos-13
+    steps:
+      - name: Checkout
+        uses: actions/checkout@v4
+      - name: SwiftLint
+        run: swiftlint --quiet --reporter github-actions-logging
```
rizwankce, line 16: This is nice feature from swiftlint itself to run it in quite mode with default reporter to make automatic comments from GitHub PR page based on linting errors

rizwankce (commented):
I made some self review comments on why I choose to do this in each line. Please review

.github/workflows/swiftlint.yml
mikaelacaron: I did this in one of my [old projects](https://github.com/mikaelacaron/brain-marks/blob/dev/.github/workflows/main.yml) and just mentioned the branch name, is there a benefit to specifically listing Swift files, vs just the branch name?

mikaelacaron (commented):
Just a question on this! Otherwise everything looks good!

.github/workflows/swiftlint.yml
mikaelacaron: @rizwankce I'm still going to merge this cause I want to see the result, but I'd like still to know the answer to that question, if you know it, thanks!

mikaelacaron: After merging I had to change the script, it does look like it was only working when manually clicking the button, and not on a PR, with Swift files changed.
I did also change it do run when commits are pushed to `main` or `dev`

rizwankce: Oh wow. Lots of things happened while I sleep. I'm from the other side of the world. 

To answer your question, if we target based on branch name. Job will run even for documentation changes or any other settings changes. That's why we only prefer to run when config files changes or any swift file changes. We are not gonna lint any other files with swift lint. Generally Mac OS machines are costly in CI. So typically we try to save the extra job which is going to be wasted. Other than that I don't find any other reason. If it works we can keep it as it is.

mikaelacaron: haha the 1st week of Hacktoberfest is always the busiest!
That's a good point! I'll keep it like this for right now, because it only runs on PRs (when I approve of it) so it should be fine if I start hitting limits, I'll change it back to the `path`
Thanks for the insight!

rizwankce: You won't be hitting any limits. GitHub actions for OSS projects is unlimited. Even for Mac OS.

mikaelacaron: Oh! I thought it was still capped, good to know!! Thanks!

PR #74, opened 2023-10-04
Issue #11 - Add error handling for addEvent method in DashboardViewModel

### LINKED ISSUE #11: IMPROVE - Error Handling When Creating an Event

achi113s:
These changes add error handling for the addEvent(_ maintenanceEvent: MaintenanceEvent) method in DashboardViewModel. They also modify DashboardView and AddMaintenanceView to appropriately present an alert to the user when an exception occurs without dismissing the AddMaintenanceView (see notes). The AddMaintenanceView is dismissed if no error is thrown.

# What it Does
* Closes #11 
* Add a do-catch block to addEvent(_ maintenanceEvent: MaintenanceEvent).
* Changes present of AddMaintenanceView from a sheet to a pushed view with the NavigationStack.

# How I Tested
* Throw a sample error from the do block; I used a BadURL error.
For example:
* Run the application
* Tap the (+) button
* Tap the Add button from the AddMaintenanceEvent screen

# Notes
* Adding an alert to this view is tricky for a few reasons. 
  * First, an alert cannot be shown over a sheet since the DashboardView can only present one view at a time. So we need to change the presentation of AddMaintenanceEvent from a sheet to a view pushed onto DashboardView with the NavigationStack.
  * Second, the alert modifier then needs to be attached to the AddMaintenanceEvent. Trying to use the existing alert for the event deletion error handling results in a warning that we are presenting from a detached view. However this then results in having two alert modifiers and needing a separate variable for presenting each one.
  * Third, there seems to be a bug in SwiftUI that results in a UILayoutConstraint warning when the alert is dismissed.
* At this point, maybe controlling the AddMaintenanceEvent view from the DashboardViewModel is not sufficient and it needs its own small view model?
* Even though this turned out to be more tricky than I thought, I will persist until we get it right lol.

# Screenshot
Screenshot 2023-10-02 at 4 52 38 PM
Screenshot 2023-10-04 at 4 23 39 PM

### DIFF
--- a/Basic-Car-Maintenance/Shared/Dashboard/ViewModels/DashboardViewModel.swift
+++ b/Basic-Car-Maintenance/Shared/Dashboard/ViewModels/DashboardViewModel.swift
@@ -13,7 +13,10 @@ import Foundation
 class DashboardViewModel: ObservableObject {
     let authenticationViewModel: AuthenticationViewModel
     
+    @Published var showAddView = false
+    
     @Published var events = [MaintenanceEvent]()
+    @Published var showAddErrorAlert = false
     @Published var showErrorAlert = false
     @Published var errorMessage : String = ""
     @Published var sortOption: SortOption = .custom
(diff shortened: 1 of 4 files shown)


Basic-Car-Maintenance/Shared/Dashboard/ViewModels/DashboardViewModel.swift
```diff
@@ -13,7 +13,10 @@ import Foundation
 class DashboardViewModel: ObservableObject {
     let authenticationViewModel: AuthenticationViewModel
     
+    @Published var showAddView = false
```
mikaelacaron, line 16: Using this boolean is one way to do this, but another is:
```swift
NavigationLink {
    AddMaintenanceView() { event in
        Task {
            await viewModel.addEvent(event)
        }
    }
} label: {
    Image(systemName: "plus")
}
```

mikaelacaron (approved):
Great work! I also made a comment, but it's not critical
Yes typically you can use `NavigationPath`, but this is only really needed when we want to programmatically control the navigation (or if you navigate several screens deep, which this doesn't so it's fine

Basic-Car-Maintenance/Shared/Dashboard/ViewModels/DashboardViewModel.swift
achi113s: This makes sense and simplifies things by allowing us to delete the boolean! I was under the impression that since we were using a NavigationStack I had to use NavigationPath. Guess not! Thanks!

mikaelacaron: @achi113s Turns out you do need to use this boolean! haha I couldn't figure out how to pop the view after the event was added

achi113s: hahaha its always the smallest features that cause the most headaches

Issue #67, opened 2023-10-04
IMPROVE - configs for widget signing


mattmassicotte:
## Describe the feature you want to improve

Signing of the newly added Widget target doesn't use the user xcconfig file. This is because the user config applies only to a single target (the app) and not the project as a whole.

A solution, though definitely not the only, is to move to an arrangement where the user-specific information (currently in Basic-Car-Maintenance.xcconfig) is imported into the project-wide xcconfig. This will allow you to apply the same signing information to all targets, which is almost certainly what you want.

mikaelacaron:
Fixed with #86

PR #77, opened 2023-10-05
Feature/setup improvements

elmoritz:
# What it Does
There is no issue for this and this would be more of a start for the improved project setup

There are now more than one xcconfig-file, because the setup does not work for the extension.
I stayed with the setup process that @mikaelacaron provided, because it is simple and easy to get behind.

With the extra xcconfig files it is possible to set the correct Bundle Identifiers for all future Apps, Extension, Targets, ...

I also added the testplans for UI-Tests and Unit-Tests, so that we can work with those in the CI.
this should be a big improvement.

I do not expect this PR to be merged like this.

There is a lot going on and I'm open for Ideas an Improvements.

As far as I can tell, with this setup improvements and Mikaelas explanation in the CONTRIBUTOR.md this should be very easy for every one to start getting into this project.

The Setup should be tested by someone not that familiar with the Project to check if I did not forget anything.

As for the Testplan:
It will be updated automatically and can also be used to setup different Configurations

image


elmoritz:
@mikaelacaron had some time left over and thought I might be of help 😇
I hope I do not interrupt the whole workflow with issues and pull request that much, just want to make it easier for everyone to get startet with this very nice little open source project

mikaelacaron:
Thanks so much @elmoritz ! Yeah I know I messed up the config with adding the widget target haha and I totally forgot about this before pushing the change yesterday #67 was looking at fixing this, he helped with the initial set up and the first xcconfig.

mikaelacaron:
@elmoritz Does the test plans work with fastlane?

elmoritz:
@mikaelacaron the testplan should make it easier for any ci tool, including fastlane.
That is also why I added 2 testplans one for unit- and one for UI-tests.

elmoritz:
You can just use the testplans from this PR and commit it.
I guess everything else would be covered by open issues

elmoritz:
> @elmoritz Does the test plans work with fastlane?

as you can see [here](https://docs.fastlane.tools/actions/scan/) in the fastlane docs you can define a testplan very easily

elmoritz:
> Thanks so much @elmoritz ! Yeah I know I messed up the config with adding the widget target haha and I totally forgot about this before pushing the change yesterday #67 was looking at fixing this, he helped with the initial set up and the first xcconfig.

My bad, I started with the testplans and ran into the building issues, so found a nice solution for it.

PR #80, opened 2023-10-06
4 feature edit maintenance event detail

### LINKED ISSUE #4: FEATURE - Edit Maintenance Event Details
* Add a trailing swipe action to edit a maintenance event that's on the `DashboardView`
* A user would swipe left, tap on the action, and a sheet will be displayed (similar to the `AddMaintenanceView`) to edit the values of the event

* One alternative is to change the `AddMaintenanceView` to update the data, as well as add it (this requires keeping track of the state of the view whether you're in "adding" mode or "updating" mode)
   * This is harder, so I'd recommend the way described above first

## Additional context

aaronthebald:
# What it Does
* Closes #4
* Using @Binding you can now swipe on an event, and press the edit button. This hands the selectedEvent over to the EditMaintenanceEventDetailsView. From there the text fields are prepopulated with the data from the selectedEvent. Once updated, using the @DocumentID to find the correct document in firebase the function overwrites the old data with the new. Then the function calls getMaintenanceEvents() to update the array of current events.

# How I Tested
* swipe on existing event.
* press edit
* update text fields in the .sheet.
* Press update.
etc...

# Notes

# Screenshot


aaronthebald:
Please let me know if there is anything you would like me to improve. Thank you so much for setting this project up!

Basic-Car-Maintenance.xcodeproj/project.pbxproj
```diff
@@ -727,7 +731,7 @@
 				CODE_SIGN_ENTITLEMENTS = "Basic-Car-Maintenance/Basic_Car_Maintenance.entitlements";
 				CODE_SIGN_STYLE = Automatic;
 				DEVELOPMENT_ASSET_PATHS = "\"Basic-Car-Maintenance/Preview Content\"";
-				DEVELOPMENT_TEAM = "";
+				DEVELOPMENT_TEAM = MW75NQ7TG5;
```
mikaelacaron:
```suggestion
				DEVELOPMENT_TEAM ="";
```

Basic-Car-Maintenance.xcodeproj/project.pbxproj
```diff
@@ -766,7 +770,7 @@
...
```
mikaelacaron:
```suggestion
				DEVELOPMENT_TEAM = "";
```

Basic-Car-Maintenance.xcodeproj/project.pbxproj
```diff
@@ -72,7 +73,7 @@
 /* Begin PBXCopyFilesBuildPhase section */
 		FFDADF8F2ACD35A200DDEF79 /* Embed Foundation Extensions */ = {
 			isa = PBXCopyFilesBuildPhase;
-			buildActionMask = 2147483647;
+			buildActionMask = 12;
```
mikaelacaron: Do you know what this is changing? If not, I'll just change it back

Basic-Car-Maintenance/Shared/Dashboard/ViewModels/DashboardViewModel.swift
```diff
@@ -78,6 +78,21 @@ class DashboardViewModel: ObservableObject {
...
+                eventToUpdate.userID = uid
+                try? Firestore
+                    .firestore()
+                    .collection("maintenance_events")
+                    .document(id)
+                    .setData(from: eventToUpdate)
+            }
+                    await self.getMaintenanceEvents()
+        }
+
```
mikaelacaron: The indentation is off for inside this function, and can you handle the error that comes back from Firebase if the `setData` doesn't work? An alert should be shown

Basic-Car-Maintenance/Shared/Dashboard/Views/DashboardView.swift
```diff
@@ -10,6 +10,10 @@ import SwiftUI
 struct DashboardView: View {
     
     @StateObject private var viewModel: DashboardViewModel
+    @State private var isShowingEditView = false
+        @State private var selectedMaintenanceEvent: MaintenanceEvent = MaintenanceEvent(
+            title: "", date: Date(), notes: ""
+        )
```
mikaelacaron: `selectedMaintenanceEvent` this should actually be optional rather than creating an object, because we're not always selecting an event

Basic-Car-Maintenance/Shared/Dashboard/Views/EditEventDetailView.swift
```diff
@@ -0,0 +1,79 @@
...
+                } header: {
+                    Text("Notes")
+                }
+            }
+            .onAppear(perform: {
+                setMaintenanceEventValues(event: selectedEvent)
+            })
+            .navigationTitle(Text("Add Maintenance"))
+            .toolbar {
+                ToolbarItem(placement: .topBarTrailing) {
```
mikaelacaron: These should be switched, the top trailing should be the update and the top leading should be the cancel

And put a single line break between the `ToolbarItem`s, so they're not right next to each other

Basic-Car-Maintenance/Shared/Dashboard/Views/EditEventDetailView.swift
```diff
@@ -0,0 +1,79 @@
...
+                    } label: {
+                        Text("Update")
+                    }
+                    .disabled(title.isEmpty)
+                }
+            }
+            
+        }
+    }
+    func setMaintenanceEventValues(event: MaintenanceEvent) {
```
mikaelacaron, line 69: There should be a single line break before this function, so that it's not right next to the body

Basic-Car-Maintenance/Shared/Dashboard/Views/EditEventDetailView.swift
```diff
@@ -0,0 +1,79 @@
...
+                }
+                ToolbarItem(placement: .topBarLeading) {
+                    Button {
+                        var event = MaintenanceEvent(title: title, date: date, notes: notes)
+                        event.id = selectedEvent.id
+                        print("button pressed")
+                        Task {
+                            await viewModel.updateEvent(event)
+                        }
+                        dismiss()
```
mikaelacaron: You don't need to create a new event object you should be able to pass `selectedEvent` directly, and given that you don't need 3 new state properties for `title`, `date`, and `notes`, you should be able to use `selectedEvent`

See below:
```swift
TextField("Title", text: $selectedEvent.title)
```

mikaelacaron: And the `dismiss` should be inside of the `Task` so that the view doesn't dismiss if there was an error

Basic-Car-Maintenance/Shared/Dashboard/Views/EditEventDetailView.swift
```diff
@@ -0,0 +1,79 @@
...
+//
+
+import SwiftUI
+
+struct EditMaintenanceEventView: View {
+    @Binding var selectedEvent: MaintenanceEvent
+    @ObservedObject var viewModel: DashboardViewModel
+    @State private var title = ""
+    @State private var date = Date()
+    @State private var notes = ""
```
mikaelacaron, line 15: We don't need these, see other comment below

Basic-Car-Maintenance/Shared/Dashboard/Views/EditEventDetailView.swift
```diff
@@ -0,0 +1,79 @@
...
+    @Environment(\.dismiss) var dismiss
+    
+    var body: some View {
+        NavigationStack {
+            Form {
+                Section {
+                    TextField("Title",
+                              text: $title)
```
mikaelacaron: This should be on one line

Basic-Car-Maintenance/Shared/Dashboard/Views/EditEventDetailView.swift
```diff
@@ -0,0 +1,79 @@
+//
+//  EditEventDetailView.swift
```
mikaelacaron: This comment should match the view's name `EditMaintenanceEventView`

mikaelacaron (changes requested):
Great work!! I have a few changes about indentation and line breaks, also a different way to do the editing on the `EditMaintenanceEventView`

Please make these changes and click "Resolve" on the ones you fix, or if you have questions about those, and once you're done, click the re-review button

Re-review button

Basic-Car-Maintenance.xcodeproj/project.pbxproj
mikaelacaron: Click these "commit suggestion" buttons after you make your changes, it's someone with the project that I messed up and will be fixing today

Basic-Car-Maintenance/Shared/Dashboard/Views/EditEventDetailView.swift
aaronthebald: First of all, thank you so much for the detailed review! I couldn't watch live but I just rewatched the stream.

As for this issue, when I attempted to update the selectedEvent values directly, I ran into errors because the MaintenanceEvent properties are let constants. So rather than refactoring the model, I created a new object with the updated values and then in firebase overwrote the old object with the new one with the updated values.

Is there another approach I should use instead? 

I appreciate all your feedback.

Basic-Car-Maintenance.xcodeproj/project.pbxproj
aaronthebald: I have no idea. I'm not sure how it even was changed.

Basic-Car-Maintenance/Shared/Dashboard/Views/EditEventDetailView.swift
mikaelacaron: I'd say go ahead and change the model 👍

Basic-Car-Maintenance/Shared/Dashboard/ViewModels/DashboardViewModel.swift
```diff
@@ -78,6 +78,26 @@ class DashboardViewModel: ObservableObject {
...
+                    .firestore()
+                    .collection("maintenance_events")
+                    .document(id)
+                    .setData(from: eventToUpdate)
+            } catch {
+                showAddErrorAlert.toggle()
+                errorMessage = error.localizedDescription
+            }
+        }
+        await self.getMaintenanceEvents()
```
mikaelacaron, line 98: This could actually be moved into the do catch, because if there was an error, there's no need to fetch new data, because nothing would be new

Basic-Car-Maintenance/Shared/Dashboard/Views/DashboardView.swift
```diff
@@ -36,6 +38,20 @@ struct DashboardView: View {
...
                         }
+                        
+                        Button {
+                            selectedMaintenanceEvent = event
+                            isShowingEditView = true
+                        } label: {
+                            VStack {
+                                Text("Edit")
+                                Image(systemName: "pencil")
+                            }
```
mikaelacaron, line 49: You can achieve the same thing with a `Label`, and it's better accessibility wise

`Label("Edit", systemImage: "pencil")`

Basic-Car-Maintenance/Shared/Dashboard/Views/DashboardView.swift
```diff
@@ -36,6 +38,20 @@ struct DashboardView: View {
...
+                        }
+                    }
+                    .sheet(isPresented: $isShowingEditView) {
```
mikaelacaron, line 52: Just a note the sheet can be moved on the same level of modifiers as the main stuff (it doesn't have to specifically be here

Basic-Car-Maintenance.xcodeproj/project.pbxproj
```diff
@@ -71,7 +72,7 @@
...
```
mikaelacaron:
```suggestion
			buildActionMask = 2147483647;
```

mikaelacaron (approved):
Great work!! I'm going to go ahead and merge this, all the other comments I made are just side notes

aaronthebald:
Thank you! Looking forward to the rest of the month

PR #112, opened 2023-10-09
Add option to cache SPM packages

### LINKED ISSUE #81: IMPROVE - Build time for docc generation with Github Action

As of now it takes nearly [5 mins](https://github.com/mikaelacaron/Basic-Car-Maintenance/actions/workflows/docc.yml) or more for building documentation for our project. Majority of the time spends on downloading the SPM packages and linking. If we could cache there files (which are not changed often), we could reduce the time taken for the whole build.

The idea is to cache the SPM packages which are downloaded and reuse the same for continuous builds. Specify a cache key based on `Package.resolved` file. Official Github Action cache [documentation](https://github.com/actions/cache/blob/main/examples.md#swift---swift-package-manager) provides a easy way to do it. But it might not work for our situation because we provide custom derived path. However cache can also have custom path.

None

rizwankce:
# What it Does
* Closes #81 
* To make a documentation CI job run faster by caching the derived data folder along with build folder

# How I Tested
* Tested on my fork repo. Please check the [link](https://github.com/rizwankce/Basic-Car-Maintenance/actions/workflows/docc.yml) for the difference

# Notes
* Caching was able to save some time for the each build. For the first time since there is no cache it will run on usual time but from the second time onwards it will be little faster. 
* first run without cache takes `7m 31s` and second run with cache takes `5m 16s`

# Screenshot
Screenshot 2023-10-09 at 12 24 53 PM
Screenshot 2023-10-09 at 12 25 01 PM

### DIFF
--- a/.github/workflows/docc.yml
+++ b/.github/workflows/docc.yml
@@ -25,6 +25,14 @@ jobs:
       - name: Setup Config file
         run: |
           cp Basic-Car-Maintenance.xcconfig.template Basic-Car-Maintenance.xcconfig
+      - uses: actions/cache@v3
+        with:
+          path: |
+            .build
+            .derivedData
+          key: ${{ runner.os }}-spm-${{ hashFiles('**/Package.resolved') }}
+          restore-keys: |
+            ${{ runner.os }}-spm-
       - name: Run Build Docs
         run: ./build-docc.sh
       - name: Setup Pages


rizwankce:
@mikaelacaron This PR tackles the speediness of the CI job. We don't get consistent data from the GitHub runners - but we do get 3min savings from the SPM caching.

mikaelacaron (approved):
Is there ever going to be a need to delete the cache? Should we make a manual action, for if that is ever needed?

rizwankce:
@mikaelacaron 

We don't need to delete the cache. It's automatically done on every job. However for stored cache can be deleted from GitHub UI. 

Do we need to delete it in any circumstance?

mikaelacaron:
> However for stored cache can be deleted from GitHub UI.
How do I do that?

I'm guessing if there's ever an issue with the CI, at some point we'd want to be able to delete the cache as a first troubleshooting step?

mikaelacaron:
@rizwankce I don't think the cache has made a difference for me, it's still taking 6-7 minutes to run the deploy documentation job

rizwankce:
@mikaelacaron it will work only on second time onwards. However I will check on this ropo GitHub actions.

mikaelacaron:
No worries at all, I'm letting you know, so far, I haven't seen any impact.

rizwankce:
Just take a look at it. Time saving is not as significant as I saw before on my repo. 

I'm gonna take another look at it. I'll send a PR if I found better solution.

mikaelacaron:
sounds good! it's not critical so it's fine if you want to move onto different issues!

rizwankce:
@mikaelacaron its seems the performance of the CI machines is not consistent. Now I see 4 mins job on this repo. In that only 2 mins for building the doc. It could be based on the Github Free plan. Im not sure. 

So I will move with other issues.

PR #114, opened 2023-10-09
101 update german localization

### LINKED ISSUE #101: Update German Localization
* Since the German localization was added to the project new strings have been added
* Update to 100% coverage

* N/A

## Additional context
* Currently at 64% coverage at the time of creating the issue

appfrosch:
# What it Does
* Closes #101 
* Updates missing german localization for new strings

# How I Tested
* Run the application

# Notes
There are a few anglicisms that I consider accepted in the german language–which is why I did not translate them but left them as they are. A good example for this is "Dashboard" which is a term used in German as well.


mikaelacaron (approved):
Great work!! Also another thing would be to download SwiftLint, and correct the warnings too

PR #119, opened 2023-10-09
#5 Edit Vehicle Details

### LINKED ISSUE #5: FEATURE - Edit Vehicle Details
* Add a trailing swipe action to edit a vehcile's details that's on the `SettingsView`
* A user would swipe left, tap on the action, and a sheet will be displayed (similar to the `AddVehicleView`) to edit the values of the vehicle

* One alternative is to change the `AddVehicleView` to update the data, as well as add it (this requires keeping track of the state of the view whether you're in "adding" mode or "updating" mode)
   * The first way is easier, so I'd recommend doing that first

## Additional context

TimothyALewis:
# What it Does
* Closes #5
* Adds EditVehicleView to spec and EditVehicle ViewModel that updates changes to Firebase


mikaelacaron (changes requested):
I DM'd you about the changes that are needed. Thanks! Once you make those changes you can push to this branch and the PR will be automatically updated.

Then click the re-review button so I'm notified.
Or if you start working on a different branch, you can close this PR and make a new one
Re-review button

mikaelacaron:
Closing, because changes need to be made, and we've DM'd about this

PR #129, opened 2023-10-10
FIX: Set upper limit for dynamic type size on DatePicker

### LINKED ISSUE #104: IMPROVE - Fix Label with Large Dynamic Type on `AddMaintenanceView`
* When the dynamic type size of the `AddMaintenanceView` is at the largest size the date label is cut off by the date picker
* When it's at this size change the label to be above the date picker (so that it isn't cut off anymore, see screenshot)

* Another alternative is to limit the dynamic type size to the largest that it can go without clipping, [see this article](https://useyourloaf.com/blog/restricting-dynamic-type-sizes/)
   * I would prefer changing the layout before limiting the dynamic type size

IMG_5494

abhinaypratap:
# What it Does
* Closes #104 
* Prevents the text in DatePicker from truncating or vanishing on larger font sizes.

# Notes
* Available font sizes:
```swift
enum DynamicTypeSize {
    case xSmall
    case small
    case medium
    case large
    case xLarge
    case xxLarge
    case xxxLarge
    case accessibility1
    case accessibility2
    case accessibility3
    case accessibility4
    case accessibility5
    // ...
}

```
* Minimum deployment of this project is set to `iOS 17.0`. The lowest resolution device compatible for `iOS 17.0` would be `iPhone SE (2nd generation)`. For that resolution, largest size that fits without breaking is `DynamicTypeSize.accessibility2`.

### DIFF
--- a/Basic-Car-Maintenance/Shared/Dashboard/Views/AddMaintenanceView.swift
+++ b/Basic-Car-Maintenance/Shared/Dashboard/Views/AddMaintenanceView.swift
@@ -35,6 +35,7 @@ struct AddMaintenanceView: View {
                     Text("Date",
                          comment: "Date picker label")
                 }
+                .dynamicTypeSize(...DynamicTypeSize.accessibility2)
                 
                 Section {
                     TextField(text: $notes,


mikaelacaron (approved):
Awesome work!!

PR #131, opened 2023-10-10
Completed initial French localization

Dean151:
# What it Does
* #7
* Added French translations to the app

# How I Tested
* Ran the application in French
* Make a tour of the application to check translations context.

# Notes
* "Make" translation key is … vague. Adding a comment to tell it's the maker of the car would be interesting.


mikaelacaron (approved):
Awesome! Thank you!

PR #134, opened 2023-10-11
Crash fix: Added document ID to vehicleToAdd

### LINKED ISSUE #117: BUG - Deleting a just recently added vehicle leads to a fatalError
## Describe the bug
* When a vehicle is added in `SettingsView` and a user tries to delete it immediately, we experience a fatal error. This is because the vehicle `id` is not set when the vehicle is added to the `vehicles` array. This is the exact location for the error source in SettingsViewModel:
```
guard let documentId = vehicle.id else {
     fatalError("Event \(vehicle.name) has no document ID.")
}
```

## To Reproduce
1. Go to the 'SettingsView` screen
2. Tap on the button 'Add Vehicle'
3. Swipe right on the just added vehicle and click 'Delete'
4. See crash

## Expected behavior
* Vehicle just added can be deleted immediately without any crash

## Screenshots
* If applicable, add screenshots to help explain your problem.

 * Device: [iPhone14 Pro]
 * OS: [e.g. iOS 17.0]
 * App Version [e.g. v1.5]

## Additional Context
* None

dogra25130:
# What it Does
* Closes #117 
* Adds document ID to vehicleToAdd before appending it to `vehicles`  array

# How I Tested
1. Run application
2. Move to settings tab
3. Add a new vehicle
4. Immediately delete the newly added vehicle

# Screen recording
https://github.com/mikaelacaron/Basic-Car-Maintenance/assets/32197474/ac28c775-5063-47cd-985d-8321d47cd632

### DIFF
--- a/Basic-Car-Maintenance/Shared/Settings/ViewModels/SettingsViewModel.swift
+++ b/Basic-Car-Maintenance/Shared/Settings/ViewModels/SettingsViewModel.swift
@@ -56,9 +56,9 @@ final class SettingsViewModel {
             
             do {
                 let firestoreRef = try Firestore
-                                        .firestore()
-                                        .collection("vehicles")
-                                        .addDocument(from: vehicleToAdd)
+                    .firestore()
+                    .collection("vehicles")
+                    .addDocument(from: vehicleToAdd)
                 vehicleToAdd.id = firestoreRef.documentID
                 vehicles.append(vehicleToAdd)
             } catch {
--- a/Basic-Car-Maintenance/Shared/Settings/ViewModels/SettingsViewModel.swift
+++ b/Basic-Car-Maintenance/Shared/Settings/ViewModels/SettingsViewModel.swift
@@ -55,11 +55,11 @@ final class SettingsViewModel {
             vehicleToAdd.userID = uid
             
             do {
-                try Firestore
-                    .firestore()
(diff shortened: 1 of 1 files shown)


Basic-Car-Maintenance/Shared/Settings/ViewModels/SettingsViewModel.swift
```diff
@@ -55,11 +55,11 @@ final class SettingsViewModel {
...
-                try Firestore
-                    .firestore()
-                    .collection("vehicles")
-                    .addDocument(from: vehicleToAdd)
-                
+                let firestoreRef = try Firestore
+                                        .firestore()
+                                        .collection("vehicles")
+                                        .addDocument(from: vehicleToAdd)
+                vehicleToAdd.id = firestoreRef.documentID
```
mikaelacaron:
```suggestion
                let firestoreRef = try Firestore
                    .firestore()
                    .collection("vehicles")
                    .addDocument(from: vehicleToAdd)
                vehicleToAdd.id = firestoreRef.documentID
```

mikaelacaron: I don't want it indented that far over because that's not the default by Xcode

mikaelacaron (approved):
Great work!!

Issue #143, opened 2023-10-11
FEATURE - Search `MaintenanceEvent`s on the `DashboardView`


mikaelacaron:
## Describe the feature you want the app to have that it currently doesn't
* Add a search bar to be able to search through the titles of the maintenance events on the `DashboardView`

## Describe alternatives you've considered
* N/A

## Additional context
* N/A

sergpryimachuk:
Hey @mikaelacaron, I can take this one 🙋‍♂️

mikaelacaron:
Closed with #150

PR #150, opened 2023-10-11
 #143 FEATURE - Search MaintenanceEvents on the DashboardView

sergpryimachuk:
# What it Does
- #143 
- Adds ability to search `MaintenanceEvent`s on the `DashboardView` through the titles.

# How I Tested
1. Run the app.
2. Add some `MaintenanceEvent`s.
3. Tap on the search bar and type a query.


Basic-Car-Maintenance/Shared/Dashboard/Views/DashboardView.swift
```diff
@@ -20,9 +20,10 @@ struct DashboardView: View {
     }
     
     var body: some View {
+        @Bindable var viewModel = viewModel
```
mikaelacaron: this doesn't need to be here, it's already defined above

Basic-Car-Maintenance/Shared/Dashboard/ViewModels/DashboardViewModel.swift
```diff
@@ -29,6 +30,14 @@ class DashboardViewModel {
         }
     }
     
+    var viewEvents: [MaintenanceEvent] {
```
mikaelacaron: can you change this variable name, instead call it `searchedEvents`

mikaelacaron (changes requested):
Great work! Just a few small changes. After you fix those, can you click the re-review button again, and I'll be notified, thanks!


Re-review button

sergpryimachuk:
Voilà, fixed! Thanks for the review 😊

mikaelacaron (approved):
Great work!!

PR #151, opened 2023-10-11
#7 FEATURE - Add initial Ukrainian localization

sergpryimachuk:
# What it Does
- #7 
- Adds initial Ukrainian localization.

# How I Tested
- Run the app in the simulator with the language set to Ukrainian.
- Check if all screens are correct.

# Notes
- Ongoing effort.
- I'll be translating new strings as the project grows.

### DIFF
--- a/Basic-Car-Maintenance/Shared/Localizable.xcstrings
+++ b/Basic-Car-Maintenance/Shared/Localizable.xcstrings
@@ -71,6 +71,12 @@
             "state" : "translated",
             "value" : "🦄 Mikaela Caron - Maintaineur"
           }
+        },
+        "uk" : {
+          "stringUnit" : {
+            "state" : "translated",
+            "value" : "🦄 Мікаела Карон - адміністраторка"
+          }
         }
       }
     },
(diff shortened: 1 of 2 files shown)


Basic-Car-Maintenance/Shared/Localizable.xcstrings
```diff
@@ -71,6 +71,12 @@
             "state" : "translated",
             "value" : "🦄 Mikaela Caron - Maintaineur"
           }
+        },
+        "uk" : {
+          "stringUnit" : {
+            "state" : "translated",
+            "value" : "🦄 Мікаела Карон - адміністраторка"
```
mikaelacaron, line 78: I'm surprised you would translate my name too, is that common?

mikaelacaron (approved):
Great work!! I just have a question about one translation

Basic-Car-Maintenance/Shared/Localizable.xcstrings
sergpryimachuk: Yes, usually real names are transliterated into the Cyrillic script. Unlike usernames or handles.

PR #156, opened 2023-10-12
17 Alternate App Icons

### LINKED ISSUE #17: FEATURE - Alternative App Icons
## Design
* Design one or two alternative app icons, size them 1024 x 1024, with no corner radius (that happens automatically)

* Comment if you want to implement the design in code, (and Mikaela will create a new issue), this issue is only for the design
* I'm also open to a complete redesign, if you want!

## Additional context
* N/A

DandyLyons:
# What it Does
* Closes #17
  * Adds a ChooseAppIconView to SettingsView, where the user can pick an alternative app icon. 

# How I Tested
- Run the app. 
- Open Settings Tab
- Tap Change App Icon
- Choose an App Icon
- Go to Home Screen to verify that the App Icon changed. 

# Screenshot
- Screenshot 2023-10-11 at 7 07 24 PM

### DIFF
--- a/Basic-Car-Maintenance.xcodeproj/project.pbxproj
+++ b/Basic-Car-Maintenance.xcodeproj/project.pbxproj
@@ -8,6 +8,7 @@
 
 /* Begin PBXBuildFile section */
 		023057F22ACFAD79006C5A73 /* EditEventDetailView.swift in Sources */ = {isa = PBXBuildFile; fileRef = 023057F12ACFAD79006C5A73 /* EditEventDetailView.swift */; };
+		8014A4CD2AD755C8005B51F6 /* Basic-Car-Maintenance.xcconfig in Resources */ = {isa = PBXBuildFile; fileRef = FF098EFA2AB3424E003EC0FE /* Basic-Car-Maintenance.xcconfig */; };
 		898009792AD1899700604E7C /* ContributorTests.swift in Sources */ = {isa = PBXBuildFile; fileRef = 898009782AD1899700604E7C /* ContributorTests.swift */; };
 		8AEE816F2ACF37F800FC0C2A /* Action.swift in Sources */ = {isa = PBXBuildFile; fileRef = 8AEE816E2ACF37F800FC0C2A /* Action.swift */; };
 		8AEE81722ACF384D00FC0C2A /* MainTabView.swift in Sources */ = {isa = PBXBuildFile; fileRef = 8AEE81712ACF384D00FC0C2A /* MainTabView.swift */; };
@@ -537,6 +538,7 @@
 			buildActionMask = 2147483647;
 			files = (
 				FF5D13AF2A86C2D800BC9BD6 /* Preview Assets.xcassets in Resources */,
+				8014A4CD2AD755C8005B51F6 /* Basic-Car-Maintenance.xcconfig in Resources */,
 				FFC8CDA42AA385E800D129A6 /* GoogleService-Info.plist in Resources */,
 				FF5D13AB2A86C2D800BC9BD6 /* Assets.xcassets in Resources */,
 				FF755B432A90915E00F49A13 /* Localizable.xcstrings in Resources */,
@@ -782,6 +784,7 @@
 		};
 		FF5D13C92A86C2D800BC9BD6 /* Debug */ = {
 			isa = XCBuildConfiguration;
+			baseConfigurationReference = FF098EFA2AB3424E003EC0FE /* Basic-Car-Maintenance.xcconfig */;
 			buildSettings = {
 				ALWAYS_EMBED_SWIFT_STANDARD_LIBRARIES = YES;
 				ASSETCATALOG_COMPILER_APPICON_NAME = AppIcon;
--- a/Basic-Car-Maintenance.xcodeproj/project.pbxproj
+++ b/Basic-Car-Maintenance.xcodeproj/project.pbxproj
@@ -11,6 +11,7 @@
 		8014A4CD2AD755C8005B51F6 /* Basic-Car-Maintenance.xcconfig in Resources */ = {isa = PBXBuildFile; fileRef = FF098EFA2AB3424E003EC0FE /* Basic-Car-Maintenance.xcconfig */; };
 		8014A4CF2AD75928005B51F6 /* AppIcon.swift in Sources */ = {isa = PBXBuildFile; fileRef = 8014A4CE2AD75928005B51F6 /* AppIcon.swift */; };
 		8014A4D12AD76034005B51F6 /* ChooseAppIconView.swift in Sources */ = {isa = PBXBuildFile; fileRef = 8014A4D02AD76034005B51F6 /* ChooseAppIconView.swift */; };
+		8014A4D32AD77C92005B51F6 /* ChooseAppIconViewModel.swift in Sources */ = {isa = PBXBuildFile; fileRef = 8014A4D22AD77C92005B51F6 /* ChooseAppIconViewModel.swift */; };
 		898009792AD1899700604E7C /* ContributorTests.swift in Sources */ = {isa = PBXBuildFile; fileRef = 898009782AD1899700604E7C /* ContributorTests.swift */; };
 		8AEE816F2ACF37F800FC0C2A /* Action.swift in Sources */ = {isa = PBXBuildFile; fileRef = 8AEE816E2ACF37F800FC0C2A /* Action.swift */; };
 		8AEE81722ACF384D00FC0C2A /* MainTabView.swift in Sources */ = {isa = PBXBuildFile; fileRef = 8AEE81712ACF384D00FC0C2A /* MainTabView.swift */; };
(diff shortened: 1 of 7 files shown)


Basic-Car-Maintenance.xcodeproj/project.pbxproj
```diff
@@ -8,6 +8,10 @@
 
 /* Begin PBXBuildFile section */
 		023057F22ACFAD79006C5A73 /* EditEventDetailView.swift in Sources */ = {isa = PBXBuildFile; fileRef = 023057F12ACFAD79006C5A73 /* EditEventDetailView.swift */; };
+		8014A4CD2AD755C8005B51F6 /* Basic-Car-Maintenance.xcconfig in Resources */ = {isa = PBXBuildFile; fileRef = FF098EFA2AB3424E003EC0FE /* Basic-Car-Maintenance.xcconfig */; };
```
mikaelacaron: It looks like you committed the `Basic-Car-Maintenance.xcconfig` file when you shouldn't have. See the [README.md](https://github.com/mikaelacaron/Basic-Car-Maintenance#getting-started) for instructions on how to copy this file properly

Basic-Car-Maintenance/Shared/Models/AppIcon.swift
```diff
@@ -0,0 +1,45 @@
...
+    case .carBlack:
+      return "Black Car"
+    case .carOrange:
+      return "Orange Car"
+    }
+  }
+  
+  var preview: UIImage {
+    UIImage(named: rawValue) ?? UIImage()
+  }
```
mikaelacaron: Is there a specific reason this has to be a `UIImage` rather than an `Image`?

Basic-Car-Maintenance/Shared/Settings/ViewModels/ChooseAppIconViewModel.swift
```diff
@@ -0,0 +1,40 @@
+import UIKit
```
mikaelacaron, line 8: You deleted the header comments, as copied from the `ChooseAppIconView`, go ahead and keep them there, cause they're on all the other files right now
```swift
//
//  ChooseAppIconView.swift
//  Basic-Car-Maintenance
//
//  Created by Daniel Lyons on 10/11/23.
```

Basic-Car-Maintenance/Shared/Settings/ViewModels/ChooseAppIconViewModel.swift
```diff
@@ -0,0 +1,40 @@
...
+    
+    Task { @MainActor in
+      guard UIApplication.shared.alternateIconName != icon.iconName else {
+        print("No need to update icon since we're already using this icon.")
+        return
+      }
+      
+      do {
+        try await UIApplication.shared.setAlternateIconName(icon.iconName)
+        print("Successfully updated icon.")
```
mikaelacaron: Can you remove the successful print statement please! It's fine to leave the "error" ones

And remove the one from line `23`, about not needing to update because that icon is already chosen. One thing you could add here is an error haptic, to still give the user feedback

Basic-Car-Maintenance/Shared/Settings/ViewModels/SettingsViewModel.swift
```diff
@@ -13,11 +13,13 @@ import Foundation
 final class SettingsViewModel {
     let authenticationViewModel: AuthenticationViewModel
     var contributors: [Contributor]?
+  var chooseAppIconViewModel: ChooseAppIconViewModel
```
mikaelacaron: Is this needed? It's not being used in the `SettingsViewModel`
I think it's fine to pass it directly when you're in the `SettingsView`

Basic-Car-Maintenance/Shared/Settings/Views/ChooseAppIconView.swift
```diff
@@ -0,0 +1,74 @@
...
+  
+  private func isSelected(_ icon: AppIcon) -> Bool {
+    return viewModel.selectedAppIcon == icon
+  }
+  
+  var body: some View {
+    let columns = [
+      GridItem(.flexible(minimum: 30, maximum: 300)),
+      GridItem(.flexible(minimum: 30, maximum: 300))
+    ]
```
mikaelacaron: You can move this outside the body, and make it a private variable

Basic-Car-Maintenance/Shared/Settings/Views/ChooseAppIconView.swift
```diff
@@ -0,0 +1,74 @@
...
+    
+    return Form {
```
mikaelacaron: When fixing the comment above, you can then remove this `return`

Basic-Car-Maintenance/Shared/Settings/Views/ChooseAppIconView.swift
```diff
@@ -0,0 +1,74 @@
...
+    
+    var body: some View {
+      GroupBox {
+        HStack {
+          Image(systemName: checkmarkImage)
+            .foregroundStyle(.blue)
+          Text(icon.description)
+          Spacer()
+        }
+        Image(uiImage: icon.preview)
```
mikaelacaron: Can you round the corners of all these previews so they look more like they would on the home screen?

mikaelacaron (changes requested):
Please fix the indentation on all of the files, it should be `4` spaces rather than `2`

Great work!! I have several comments that need resolved please. Once you fix them, please click the re-review button:
Re-review button


Also see how in the description when you write `Closes #17` it automatically links the issue. You replace the `#issueNumber` part from the template. (I edited this, but for future reference)

Basic-Car-Maintenance/Shared/Settings/Views/ChooseAppIconView.swift
DandyLyons: Good idea! Done!

Basic-Car-Maintenance/Shared/Settings/Views/ChooseAppIconView.swift
DandyLyons: Done.

Basic-Car-Maintenance/Shared/Settings/Views/ChooseAppIconView.swift
DandyLyons: Done.

Basic-Car-Maintenance/Shared/Settings/ViewModels/SettingsViewModel.swift
DandyLyons: Done.

Basic-Car-Maintenance/Shared/Settings/ViewModels/ChooseAppIconViewModel.swift
DandyLyons: Good idea. Done.

Basic-Car-Maintenance/Shared/Models/AppIcon.swift
DandyLyons: Because the image is stored in a .appIconSet, the images did not load using a plain old SwiftUI String initialization like `Image(icon.rawValue)`. (See  ChooseAppIconView (line 64).) All of the implementations that I saw so far, including the SwiftLee article you sent (which was very helpful by the way 👍🏼) used the UIImage method. 

So far I haven't found a purely SwiftUI way to do this. I could do more research on it, but I'm not sure that there's a major benefit to it.

Basic-Car-Maintenance/Shared/Settings/ViewModels/ChooseAppIconViewModel.swift
DandyLyons: 👍🏼

Basic-Car-Maintenance.xcodeproj/project.pbxproj
DandyLyons: I must have misunderstood something. I did run `cp Basic-Car-Maintenance.xcconfig.template Basic-Car-Maintenance.xcconfig` like the README says. 

The file itself is not in my commit history. I guess you mean that I shouldn't have added `Basic-Car-Maintenance.xcconfig` to the .xcodeproj?

Is that the reason why you had `#include "Basic-Car-Maintenance.xcconfig"` in the `Project.xcconfig`. If so, I think something must be set up wrong on my machine because the `#include` wasn't finding the file, and the project wasn't building until I added the `Basic-Car-Maintenance.xcconfig` to the project. 

However it does appear to be building now (even after removing `Basic-Car-Maintenance.xcconfig`) so I guess Xcode was building based off of a stale index.

Basic-Car-Maintenance/Shared/Models/AppIcon.swift
mikaelacaron: all good, I was curious!

Basic-Car-Maintenance.xcodeproj/project.pbxproj
mikaelacaron: You should run `cp Basic-Car-Maintenance.xcconfig.template Basic-Car-Maintenance.xcconfig` while being in the root directory (where the xcodeproj is), cause  `Basic-Car-Maintenance.xcconfig` is in the `.gitignore`

The include is to override values from the main config and also use them in the Project.xcconfig

And you run that command, before opening Xcode!

DandyLyons: Ok it sounds like the issue was that I opened Xcode before running the command. Lesson learned.

mikaelacaron: I'll add this to the README for more clarification 👍

Basic-Car-Maintenance.xcodeproj/project.pbxproj
```diff
@@ -100,7 +106,6 @@
 		E58499692ACDDAFF00634660 /* Contributor.swift */ = {isa = PBXFileReference; lastKnownFileType = sourcecode.swift; path = Contributor.swift; sourceTree = "<group>"; };
 		FF0813562AD0A83000910EFA /* UITests.xcconfig */ = {isa = PBXFileReference; lastKnownFileType = text.xcconfig; path = UITests.xcconfig; sourceTree = "<group>"; };
 		FF0813572AD0A92700910EFA /* Widget.xcconfig */ = {isa = PBXFileReference; lastKnownFileType = text.xcconfig; path = Widget.xcconfig; sourceTree = "<group>"; };
-		FF098EFA2AB3424E003EC0FE /* Basic-Car-Maintenance.xcconfig */ = {isa = PBXFileReference; lastKnownFileType = text.xcconfig; path = "Basic-Car-Maintenance.xcconfig"; sourceTree = SOURCE_ROOT; };
```
mikaelacaron: @DandyLyons Can you go through and make sure this is added here?
Sorry I misspoke earlier the `Basic-Car-Maintenance.xcconfig` is ignored in the `.gitignore`, but it's still used in Xcode, because that's how it's used in the project

I think you should be able to use any text editor, and add these lines back, just make sure they're in the right place

mikaelacaron (changes requested):

Basic-Car-Maintenance.xcodeproj/project.pbxproj
DandyLyons: Done.

mikaelacaron (approved):
Awesome!! Looks great!
Also note I fixed the `pbxproj` file, I manually edited the changes in VSCode, for stuff that was changed related to the `Basic-Car-Maintenance.xcconfig`

PR #161, opened 2023-10-12
Fix event deletion

### LINKED ISSUE #63: BUG - Delete Maintenance Event Not Removed From Array
## Describe the bug
* The maintenance event does the swipe to delete, but it actually isn't being removed from the `events` array and

## To Reproduce
* Make sure you have an event created
1. Swipe left to delete the event
2. See the error, the event animates like it deletes, but then is still there

* Switching tabs and coming back to Dashboard, you can see that it is deleted from Firebase, but it doesn't remove from the array, staying on the `DashboardView`

## Expected behavior
* Swipe to delete, the event is removed from the `events` array and from Firebase

## Screenshots

https://github.com/mikaelacaron/Basic-Car-Maintenance/assets/22946902/be9b9c4b-9f86-451f-9a00-f2eebcb01f1b

 * OS: iOS 17.0
 * App Version v1.0 (1)

josephnoir:
# What it Does
* Closes #63
* When a maintenance event is deleted it is now removed from the local model as well.

# How I Tested
* Run the application
* Tap the (+) button and add a new event
* Swipe the event to the left
* Click the `Delete` button
* The event now disappears from the list

# Screenrecording

https://github.com/mikaelacaron/Basic-Car-Maintenance/assets/2156518/a8c135d4-c596-4854-971d-1383370a2102

### DIFF
--- a/Basic-Car-Maintenance/Shared/Dashboard/ViewModels/DashboardViewModel.swift
+++ b/Basic-Car-Maintenance/Shared/Dashboard/ViewModels/DashboardViewModel.swift
@@ -111,6 +111,10 @@ class DashboardViewModel {
                 .document(documentId)
                 .delete()
             errorMessage = ""
+            
+            if let eventIndex = events.firstIndex(of: event) {
+                events.remove(at: eventIndex)
+            }
         } catch {
             showErrorAlert.toggle()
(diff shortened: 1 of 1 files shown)


mikaelacaron (approved):
Awesome work!!

PR #164, opened 2023-10-12
Fix:Firebase Warning - Implemented Event Sync function for sync with Firebase after additions or deletions.

### LINKED ISSUE #159: BUG - Firebase Warning
## Describe the bug
* When you add a new `Vehicle` or `MaintenanceEvent` a warning like this is shown in the console.
* We are fetching an array of items returned from Firebase, when a new item is added, it doesn't have a document ID because it's assigned by Firebase. This creates an issue when going to delete an item immediately after it's created. See #134 the solution to this is setting the document ID manually

* But that solution, leads to this output, that Firebase doesn't want us to do this.

```
Attempting to initialize or set a @DocumentID property with a non-nil value: "ugfj380dZQf0rubpWjAL". The document ID is managed by Firestore and any initialized or set value will be ignored. The ID is automatically set when reading from Firestore.
```

### Possible Solution
One way to solve this is going back and fetching the new items to refresh the local array (and removing the statements explicitly setting the document ID

## To Reproduce
1. Go to `DashboardView`
2. Tap the `+` button
3. Fill in the text fields and tap Add
4. See the console output like above

## Expected behavior
* This output is not shown in the console
* You can create new items (`MaintenanceEvent` or `Vehicle`) and delete them right after they're created

## Screenshots
* N/A

 * Device: any iPhone
 * OS: iOS 17
 * App Version: v1.0 (1)

## Additional Context
* N/A

SarathVSasikala:
# What it Does
* Closes #159 
* Implemented Event Sync function for sync with Firebase after additions or deletions.

# How I Tested
* Run the application
* Tap the (+) button
* Added new maintenance
* back to Dashboard
* Deleted the maintenance

### DIFF
--- a/Basic-Car-Maintenance/Shared/Dashboard/ViewModels/DashboardViewModel.swift
+++ b/Basic-Car-Maintenance/Shared/Dashboard/ViewModels/DashboardViewModel.swift
@@ -84,10 +84,6 @@ class DashboardViewModel {
         }
     }
     
-    func syncEvents() async {
-        await self.getMaintenanceEvents()
-    }
-    
     func updateEvent(_ maintenanceEvent: MaintenanceEvent) async {
         
         if let uid = authenticationViewModel.user?.uid {
--- a/Basic-Car-Maintenance/Shared/Dashboard/ViewModels/DashboardViewModel.swift
+++ b/Basic-Car-Maintenance/Shared/Dashboard/ViewModels/DashboardViewModel.swift
@@ -39,16 +39,12 @@ class DashboardViewModel {
             eventToAdd.userID = uid
             
             do {
-                let documentReference = try Firestore
+                try Firestore
                     .firestore()
                     .collection("maintenance_events")
                     .addDocument(from: eventToAdd)
                 
-                var event = maintenanceEvent
-                let documentId = documentReference.documentID
-                event.id = documentId
-                
-                events.append(event)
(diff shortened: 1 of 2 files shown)


Basic-Car-Maintenance/Shared/Dashboard/Views/DashboardView.swift
```diff
@@ -38,6 +38,7 @@ struct DashboardView: View {
                         Button(role: .destructive) {
                             Task {
                                 await viewModel.deleteEvent(event)
+                                await viewModel.syncEvents()
```
mikaelacaron: I don't think we need to call sync here

Basic-Car-Maintenance/Shared/Dashboard/ViewModels/DashboardViewModel.swift
```diff
@@ -79,6 +75,10 @@ class DashboardViewModel {
         }
     }
     
+    func syncEvents() async {
+        await self.getMaintenanceEvents()
+    }
+
```
mikaelacaron: I don't think it's necessary to create a new function that just calls another function in the same scope

mikaelacaron (approved):
Good work! I left a few comments about how you implemented it, but I'm going to fix these so I can still merge this right now.
Good work!

PR #168, opened 2023-10-13
Fix:Firebase Warning - added get Vehicles after Vehicle additions

### LINKED ISSUE #159: BUG - Firebase Warning

SarathVSasikala:
# What it Does
* Closes #159 
* Resolved firebase warning.
* Added get Vehicles call after Vehicle additions

# How I Tested
For example:
* Run the application
* Navigate to the settings screen
* Add new vehicle
* Removed vehicle

### DIFF
--- a/Basic-Car-Maintenance/Shared/Settings/ViewModels/SettingsViewModel.swift
+++ b/Basic-Car-Maintenance/Shared/Settings/ViewModels/SettingsViewModel.swift
@@ -56,11 +56,10 @@ final class SettingsViewModel {
             vehicleToAdd.userID = uid
             
             do {
-                let firestoreRef = try Firestore
+                try Firestore
                     .firestore()
                     .collection("vehicles")
                     .addDocument(from: vehicleToAdd)
-                vehicleToAdd.id = firestoreRef.documentID
                 vehicles.append(vehicleToAdd)
             } catch {
                 throw error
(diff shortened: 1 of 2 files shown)


mikaelacaron (approved):
Thanks for making this change!

PR #177, opened 2023-10-15
FEATURE - implement realtime alert functionality

### LINKED ISSUE #107: FEATURE - Create a Real Time Alert System
* Create a real time alert sheet, inspired from [this video](https://youtu.be/8e1Rbh1ck-U?si=_jUUW_qtkRiYP8fo)
   * Instead of being a custom alert like the video, use a sheet on the `MainTabView` that's in the `BasicCarMaintenanceApp`, using `.presentationDetents([.medium, .large])` for the sheet looking UI
   * Make it so that the sheet can't be swiped down to be dismissed, a user must click the Dismiss button
* Please follow the spec below and let @mikaelacaron know if you have any questions! (tag me in the comment, thanks)
   * Do this in multiple PRs, as shown below

* N/A

## Additional context
* Get data from Firestore, `alerts` document, which will return an array of alerts. (call the model `RealTimeAlert`)
   * Filter by alerts where `isOn` is `true` (this is something that Mikaela will manually update in Firestore)
   * Use this structure (`actionText` and `actionURL` are optional)

Screenshot 2023-10-16 at 11 27 27 PM

### UI - Creating the Alert (PR 1)
* Use tool bar items for a `Dismiss` button on the leading edge
* Layout the view as shown in the sketch below
   * Center all the elements
   * The action button should stretch the width of the screen with `24` padding on the left and right, and be at least `44` tall
   * When the user clicks on the action button, it should open Safari for whatever website is fetched for `actionURL` (that's in the Firestore response)
   * The sheet shouldn't dismiss when swiping down, a user should have to tap the `Dismiss` button

### Functionality (PR 2)
* On every launch of the app check if there's a message with `isOn` is `true` (only fetch where `isOn` is `true`, don't fetch all of the messages)
   * If you get a message back, show the returned alert on the `MainTabView` (this means it will be presented no matter what)
      * Save the `id` of the message to an array in user defaults  (AppStorage) called `readRealTimeAlerts`
         * Don't show the alert if the `id` of that alert has already been saved to `readRealTimeAlerts` (this is so the alert is not shown every time the user opens the app, it's only shown to them once)
         * ⚠️ I'm thinking save this to user defaults or using SwiftData, but does SwiftData really add any dependencies (like how CoreData did)? let me know in the comments!
   * If you don't get anything back, don't show an alert

### Design
   * Make sure that the alert can scroll
iPhone 14 Pro - 1

OmarHegazy93:
# What it Does
Wire the implemented UI for realtime alert system with the server so that it gets updated each time new alert is arrived

* Closes #107
Add functionality to realtime alert system by:
1. listening to the server for newly added alerts
2. Add a query to get only **ONE** alert with `isOn: true`, and if there are viewed alerts previously, append to the query to exclude all alerts that ids match the persisted ids
3. if the alert is new, present it and save its id locally so that it gets filtered next time when new alerts are received

# How I Tested
- Install the app and run it for the first time
- An alert should appear on the screen
- Dismiss it
- Reopen the app
- Notice the behavior

# Notes
- When running the app for the second time, there will be an error from the firebase in the console as shown in the screenshot below, as per the answer [here](https://stackoverflow.com/a/53790652), @mikaelacaron should open that link so that it gets created automatically afterwards. (please let me know if there is something else I should do about it)
Screenshot 2023-10-15 at 20 33 46

# Screenshot


Basic-Car-Maintenance/Shared/BasicCarMaintenanceApp.swift
```diff
@@ -7,16 +7,27 @@
...
 struct BasicCarMaintenanceApp: App {
     @State private var actionService = ActionService.shared
     @UIApplicationDelegateAdaptor(AppDelegate.self) var delegate
+    let container: ModelContainer
+    
+    init() {
+        do {
+            self.container = try ModelContainer(for: AcknowledgedAlert.self)
+        } catch {
+            fatalError("Failed to create ModelContainer for AcknowledgedAlert.")
```
mikaelacaron: Is it likely this ever fails? Because this, would mean the whole app can't ever be initialized if the `container` fails to initialize

Basic-Car-Maintenance/Shared/BasicCarMaintenanceApp.swift
```diff
@@ -7,16 +7,27 @@
...
+        }
+    }
     
     var body: some Scene {
         WindowGroup {
-            MainTabView()
+            MainTabView(modelContext: container.mainContext)
                 .environment(actionService)
+                .modelContainer(container)
```
mikaelacaron: Is there a reason we're passing it via the environment, and through the initializer?
Also do we not need to do something like this anymore, because it's `SwiftData`?

```swift
@Environment(ModelContainer.self) var container
```

Basic-Car-Maintenance/Shared/MainView/Views/MainTabView.swift
```diff
@@ -64,10 +64,24 @@ struct MainTabView: View {
...
+        }
+        .onChange(of: viewModel.alert) { _, newValue in
+            guard newValue != nil else { return }
+            isShowingRealTimeAlert = true
+        }
+    }
+    
+    init(modelContext: ModelContext) {
+        let mainViewModel = MainTabViewModel(modelContext: modelContext)
+        self._viewModel = .init(initialValue: mainViewModel)
```
mikaelacaron: Please move the `init` to above the body, but below the properties

Basic-Car-Maintenance/Shared/MainView/ViewModels/MainTabViewModel.swift
```diff
@@ -0,0 +1,83 @@
...
+            .collection("alerts")
+            .whereField("isOn", isEqualTo: true)
+            .limit(to: 1)
+        
+        if !acknowledgedAlerts.isEmpty {
+            query = query
+                .whereField("_id", notIn: self.acknowledgedAlerts.map(\.id) as [String])
+        }
+        
+        query.addSnapshotListener {[weak self] snapShot, error in
```
mikaelacaron: Can you change this to `snapshot` not `snapShot`

Basic-Car-Maintenance/Shared/MainView/ViewModels/MainTabViewModel.swift
```diff
@@ -0,0 +1,83 @@
...
+        self.modelContext = modelContext
+        fetchAcknowledgedAlerts()
+    }
+    
+    /// Update the UI once a new alert is sent
+    func listenToAlertsUpdates() {
+        var query = Firestore
+            .firestore()
+            .collection("alerts")
+            .whereField("isOn", isEqualTo: true)
```
mikaelacaron: Use the new `FirestoreCollection` and `FirestoreField` enums

You'll need to create new properties

Basic-Car-Maintenance/Shared/MainView/ViewModels/MainTabViewModel.swift
```diff
@@ -0,0 +1,83 @@
...
+            .limit(to: 1)
+        
+        if !acknowledgedAlerts.isEmpty {
+            query = query
+                .whereField("_id", notIn: self.acknowledgedAlerts.map(\.id) as [String])
```
mikaelacaron: Why does this need to be `_id` rather than just `id`?

Basic-Car-Maintenance/Shared/Models/AlertItem.swift
```diff
@@ -0,0 +1,45 @@
...
+    var actionTxt: String
+    var actionURL: String
+    var icon: String
+    var isOn: Bool
+    var message: String
+    var title: String
+    
+    enum CodingKeys: String, CodingKey {
+        case id
+        case actionTxt = "actionTex"
```
mikaelacaron:
```suggestion
        case actionTxt = "actionText"
```

typo, on my end, haha spell it out correctly, and I'll fix it in Firestore

Basic-Car-Maintenance/Shared/Models/AlertItem.swift
```diff
@@ -0,0 +1,45 @@
...
+//  Created by Omar Hegazy on 13/10/2023.
+//
+
+import Foundation
+import FirebaseFirestoreSwift
+import SwiftData
+
+struct AlertItem: Codable, Identifiable {
+    @DocumentID var id: String?
+    var actionTxt: String
```
mikaelacaron: Also, cause it's show short, just spell out `actionText` rather than `actionTxt`

mikaelacaron (changes requested):
I haven't fully looked at it functionality wise, if it works (and testing with Firebase, so I'm not sure yet on the question you had about that error)

But here's several suggestions to please fix, and I'll take a look again
Great work again!! You're doing awesome!

Basic-Car-Maintenance/Shared/BasicCarMaintenanceApp.swift
OmarHegazy93: Yes! if we updated the `AcknowledgedAlert` scheme or any persistence model without migration plan. So either we keep the `do` `catch` block or to make the try optional.

Let me know please which one you prefer

Basic-Car-Maintenance/Shared/MainView/ViewModels/MainTabViewModel.swift
OmarHegazy93: Because I want to filter out ids that match the `DocumentID` as described in the screenshot below, not sure if using `id` will also do the job, we will need to test it
Screenshot 2023-10-15 at 20 31 22

Basic-Car-Maintenance/Shared/BasicCarMaintenanceApp.swift
OmarHegazy93: 1.  Passing to the view hierarchy is related to automatically update the views if something is changed
2. Passing through the initializer is because if we declared `@Environment(\.modelContext) var modelContext`
in `MainTabView` and tried to pass it through the view's initializer we will have a compiler error as `self` isn't initialized yet

> [!NOTE]
> Since I'm still new to `SwiftData`, I followed the approach in this [tutorial](https://www.hackingwithswift.com/quick-start/swiftdata/why-are-swiftdata-models-created-as-classes)

Basic-Car-Maintenance/Shared/BasicCarMaintenanceApp.swift
mikaelacaron: Will we later be needing to do a database migration here? (whenever it comes to that)

Basic-Car-Maintenance/Shared/MainView/ViewModels/MainTabViewModel.swift
mikaelacaron: Gotcha, also btw I'll update the original issue, but I just changed the model so it's not a UUID, but it will be the same as the document ID (because looking at it, it makes sense to just use the same ID, because it will be unique)

Screenshot 2023-10-16 at 11 27 27 PM

Basic-Car-Maintenance/Shared/BasicCarMaintenanceApp.swift
mikaelacaron: I think you didn't link the right tutorial? the one you linked is why you use classes for the `@Model` macro

I think it only needs to be passed once.

I'm also not an expert, because none of us are, because it's so new, but I'd follow this [Apple tutorial](https://developer.apple.com/documentation/swiftdata/adding-and-editing-persistent-data-in-your-app) which passes it through the environment and doesn't actually store it in the "view model" but instead passes the context into the vm to use as a parameter

OmarHegazy93: Sorry my bad, this is the [tutorial](https://www.hackingwithswift.com/quick-start/swiftdata/how-to-use-mvvm-to-separate-swiftdata-from-your-views) I meant
I read the tutorial you mentioned and to make sure that I understand you correctly, please let me know if the following steps are what you mean:
1. you want to have `@Environment(\.modelContext) var modelContext` in `MainTabView` and not injected through the `init`
2. in the ViewModel, you want to expose both fetch and save methods to the view and whenever we need to use them we'll pass the modelContext to them as a parameter and the VM will know nothing about modelContext?

Basic-Car-Maintenance/Shared/BasicCarMaintenanceApp.swift
OmarHegazy93: So far I don't think we will because only the id is all what we want to control the alerts

mikaelacaron: I agree, I think at the moment the `id` is fine

Basic-Car-Maintenance/Shared/BasicCarMaintenanceApp.swift
OmarHegazy93: @mikaelacaron Any thoughts?

mikaelacaron: I'll look at this tonight

mikaelacaron: Okay, I've posted about this on Mastodon and Twitter, and this is a divided issue right now

Can you for now, follow the [Apple tutorials](https://developer.apple.com/documentation/swiftdata) style, even though it's not specifically "MVVM"

I think most of Apple's style is making a function in the SwiftUI view, which is fine for now

Thanks for you patience!

OmarHegazy93: I saw your posts and it seems that it was done completely wrong, Thanks for taking care of it.
As a result, both model and context are moved to the view, so please have a look and let me know if you still have any comments

Basic-Car-Maintenance/Shared/Models/AlertItem.swift
```diff
@@ -0,0 +1,45 @@
...
+    
+    enum CodingKeys: String, CodingKey {
+        case id
+        case actionText
+        case actionURL
+        case icon = "emojiIcon"
+        case isOn
+        case message
+        case title
+    }
```
mikaelacaron: leaving this as `emojiIcon` I think is the better choice here because this is the only one that's different

Basic-Car-Maintenance/Shared/MainView/ViewModels/MainTabViewModel.swift
```diff
@@ -0,0 +1,56 @@
...
+                        print("Error decoding to AlertItem: ", error)
+                        return nil
+                    }
+                }
+                .first
+            
+            if let newAlert {
+                Task { @MainActor in
+                    self.alert = newAlert
+                }
```
mikaelacaron, line 51: Is there a reason this is wrapped in a `Task`?

Basic-Car-Maintenance/Shared/MainView/ViewModels/MainTabViewModel.swift
```diff
@@ -0,0 +1,56 @@
...
+            .collection(FirestoreCollection.alerts)
+            .whereField(FirestoreField.isOn, isEqualTo: true)
+            .limit(to: 1)
+        
+        if !acknowledgedAlerts.isEmpty {
+            query = query
+                .whereField(FirestoreField.id, notIn: acknowledgedAlerts)
+        }
+        
+        query.addSnapshotListener { [weak self] snapshot, error in
```
mikaelacaron: This doesn't necessarily need to be a listener, it could just be a fetch when the app is launched

Basic-Car-Maintenance/Shared/MainView/Views/MainTabView.swift
```diff
@@ -42,11 +49,10 @@ struct MainTabView: View {
...
         .sheet(isPresented: $isShowingRealTimeAlert) {
-            AlertView()
-                .presentationDetents([.medium])
-        }
-        .onTapGesture(count: 2) {
-            isShowingRealTimeAlert = true
+            if let alert = viewModel.alert {
+                AlertView(alert: alert)
+                    .presentationDetents([.medium])
+            }
```
mikaelacaron: An even better way to do this is to use the `.sheet(item:content:)` modifier because you can track the `alert` in the `MainTabViewModel` and not need an alert object and a boolean for showing the sheet

mikaelacaron (changes requested):
Great work!! Note I pushed several changes (so you'll need to do `git pull` on your branch)
Mainly just formatting changes here and there.

Added those indexes for `alerts`

I have a few questions. Otherwise this will be good to merge soon!

And can you make sure that the title doesn't get clipped when it's longer (see screenshot) and make sure in the message when I add `\n` for a line break, that it shows up as a line break

mikaelacaron:
showing `\n\n` rather than as line breaks
Simulator Screenshot - iPhone 15 Pro - 2023-10-25 at 01 49 15

the title is clipped
Simulator Screenshot - iPhone 15 Pro - 2023-10-25 at 02 07 35

Basic-Car-Maintenance/Shared/MainView/ViewModels/MainTabViewModel.swift
OmarHegazy93: Yes, what is happening here actually is:
1. The `alert` property is defined as `@MainActor var alert: AlertItem?` so that if you tried to update that alert from outside the main actor, you gonna get a warning that alert is being updated from out side the main actor (soon in Swift 6 it will give you a compiler error)
2. To update the `alert` on the main actor, we can use `await MainActor.run(...)` but since we're not in an async context, we can use `Task` with `@MainActor` to make sure alert is updated through the proper context

Basic-Car-Maintenance/Shared/MainView/ViewModels/MainTabViewModel.swift
OmarHegazy93: I thought about it, but fetching Alerts only once when the app is launched won't make it "realtime alert", that's why I preferred to fetch the alert if you add something new instantly.
But it's ok to make it the way you want, just let me know please.

mikaelacaron: Maybe realtime isn't really the right word. It's "realtime" as in faster than the App Store haha and sent from a backend

So yeah if you could change it to just a single fetch on app launch, that'd be great!

Otherwise someone could be doing something and they would suddenly get this alert, interrupt them, they'd dismiss it, not read it, and right now, have no way to get back that message

mikaelacaron (approved):
The code looks good but I haven't been able to test it successfully with the functionality that I'm wanting.

I'm going to accept it, and look at it again after Hacktoberfest, or put it behind a feature flag for now, thanks for all your work on this!

PR #196, opened 2023-10-21
Bug-Fix issue: 182 Picker in AddMaintainanceView is not working

### LINKED ISSUE #182: BUG - Picker in AddMaintenanceView is not working
## Describe the bug
It is not possible to select another vehicle in the AddMaintenanceView.

## To Reproduce
1. Add more than on vehicle in the settings screen
2. Add a new maintenance
3. Try to select the second vehicle in the picker

## Additional Context
* Problem is that the picker selection should only use id not the vehicle object.

@mikaelacaron  I already have a working bugfix, but I wanted to give someone else the chance to fix it. If the solution is needed immediately I can open a PR with the fix.

rramirez-dev:
Updated the vehicle picker to user the vehicle id and not the vehicle object

# What it Does
* Closes #182 
   The fix allows the user to select a vehicle from the 
   vehicle picker in the AddMaintainanceView form that
   is not the first vehicle in the list.

# How I Tested
For example:
* Run the application
* Tap the (+) button
* See this screen
etc...

1. Launch the App
2. Load the dashboard by navigating to a different tab     and then select the dashboard tab again.
3. Tap the plus sign to add a new event
4. Enter a title
5. Select a vehicle that is not the first one on the list
6. Tap Add

# Notes
* I filtered the vehicles array based on the id of the vehicle that the user selected. I then get the first item in the array and pass it to the MaintainanceEvent view.

# Screenshot
Simulator Screen Recording - iPhone 15 - 2023-10-20 at 21 49 41


mikaelacaron (commented):
Can you make this work using the `Vehicle` object as opposed to the id? if not use the id and change the variable to `selectedVehicleID`
Also there's some indentation that's incorrect, your Xcode should be set to `4` spaces

mikaelacaron (changes requested):
nevermind sorry for the confusion, go ahead and use the `id`, that's how it's done in the `AddOdometerReadingView`

Also note you don't need the `vehicle as Vehicle` anymore because of that, and technically with the current way that you fixed this the `if !vehicles.isEmpty` check wouldn't be needed, because you made the `selectedVehicle` optional

rramirez-dev:
So I should remove "vehicle as Vehicle" from the tag and replace it with "vehicle.id".

mikaelacaron:
> So I should remove `vehicle as Vehicle` from the tag and replace it with `vehicle.id`

Yes because it's not necessary to cast the vehicle as it's own type, the goal is to use the `id` as the main point of data for the picker, like you had it before

Basic-Car-Maintenance/Shared/Dashboard/Views/AddMaintenanceView.swift
```diff
@@ -34,9 +34,9 @@ struct AddMaintenanceView: View {
                 }
                 
                 Section {
-                    Picker(selection: $selectedVehicle) {
-                        ForEach(vehicles) { vehicle in
-                            Text(vehicle.name).tag(vehicle as Vehicle)
+                    Picker(selection: $selectedVehicleID) {
+                        ForEach(vehicles, id: \.self) { vehicle in
```
mikaelacaron:
```suggestion
                        ForEach(vehicles) { vehicle in
```

This is not needed, it's about if the elements in the `ForEach` are `Identifiable`, which they are, the `id` doesn't need to be explicitly noted

Basic-Car-Maintenance/Shared/Dashboard/Views/AddMaintenanceView.swift
```diff
@@ -67,23 +67,21 @@ struct AddMaintenanceView: View {
...
-                                                         notes: notes,
-                                                         vehicle: selectedVehicle)
-                            addTapped(event)
+                        if let selectedVehicleID {
+                            if let vehicle = vehicles.filter({ $0.id == selectedVehicleID }).first {
+                                let event = MaintenanceEvent(title: 
+                                                            title,
+                                                            date: date,
+                                                            notes: notes,
+                                                            vehicle: vehicle)
```
mikaelacaron: Going to push a change to fix the alignment of this

mikaelacaron (approved):
Good work! Just a small comment and I can merge this

PR #220, opened 2023-10-24
`AddVehicleView` improvements

### LINKED ISSUE #208: IMPROVE - Make AddVehicleView more interactive

1. Vehicle Make field:
There's a constant number of vehicle makers, so we can add a picker to simplify filling the form.

2. Year field:
Adding a year can also be implemented with a picker.

3. VIN and licence number fields:
Those are usually uppercase, so we can add .textCase(.uppercase) view modifier to reinforce the format (user still will be able to change text case).

- N/A

sergpryimachuk:
# What it Does
* Closes #208 
* Adds keyboard related viewModifiers.

# How I Tested
* Run the app in the Simulator.
* Manually check if the keyboard behaviour is correct.

# Notes
* Very simple PR that makes me a little embarrassed 😆 

# Screenshot
- N/A


mikaelacaron (approved):
Great work! Thanks for fixing those SwiftLint warnings

Someone else also noted them in their PR but they haven't updated theirs yet

PR #223, opened 2023-10-24
Implements contribution tip

### LINKED ISSUE #214: FEATURE - Add TipKit
* Implement a TipKit tip that shows a user on the `SettingsView` that this an app that's open source and that anyone can contribute to this project if they're interested in iOS development with Swift

* N/A

## Additional context
* I haven't used TipKit yet, but I did not want to make a tip pointing to the `+` button because that feels obvious and self explanatory. If you have any other ideas for a tip, comment below what you want to implement it and @mikaelacaron will approve it (if it's different than what's described above)

SandroDahl:
# What it Does
* Closes #214 
* Implements a contribution tip for the Github repo link in settings. 

# How I Tested
* Run the application
* Open settings

# Notes
* Localizations needs to be updated!

# Screenshot


Basic-Car-Maintenance/Shared/Settings/Tips/ContributionTip.swift
```diff
@@ -0,0 +1,23 @@
...
+//
+
+import TipKit
+
+struct ContributionTip: Tip {
+    var title: Text {
+        Text("Thanks for using this app!", comment: "Thanks a user for using the app.")
+    }
+
+    var message: Text? {
```
mikaelacaron: Why is message and image optional, when they're both computed properties that return a value?

Basic-Car-Maintenance/Shared/Settings/Views/SettingsView.swift
```diff
@@ -29,53 +30,59 @@ struct SettingsView: View {
...
+                        } label: {
+                            HStack {
+                                Image(systemName: "person.3.fill")
+                                Text("Contributors", comment: "Link to contributors list.")
+                            }
+                        }
                     }
+                    .foregroundStyle(.blue)
+                } header: {
+                    Text("Contribution")
```
mikaelacaron: Can you remove this Section header (and moving the whole thing into a section)? I think it's fine without it right now, also add back this text at the top. Even though yes it'll show twice the first time the Tip shows, I think that's fine, I do want that text to always be shown.

```swift
 // swiftlint:disable:next line_length
Text("Thanks for using this app! It's open source and anyone can contribute to it.", comment: "Thanks a user for using the app and tells the user they can contribute to the codebase")
```

mikaelacaron (changes requested):
Great work!!
Can you move the Tips folder you created, move it from inside `Settings` to instead be right above the `Utilities` folder (Because right now I'd like all the tips to be at a high level, especially cause there's just one)

If you have any questions, you can comment directly on that thread, otherwise click the resolve button once you've addressed my comment. Once you fix everything, please click the re-review button

Re-review button

Basic-Car-Maintenance/Shared/Settings/Tips/ContributionTip.swift
SandroDahl: Otherwise nothing is displayed. Seems to be a TipKit bug.

Basic-Car-Maintenance/Shared/Settings/Views/SettingsView.swift
SandroDahl: Yup, I changed it.

SandroDahl:
All navigation links are not grayed out when the tip is focused. At the moment I don't have a solution for this.

mikaelacaron:
> All navigation links are not grayed out when the tip is focused. At the moment I don't have a solution for this.


And right now then the solution is to put it in a section with a header, and that fixes it? @SandroDahl

SandroDahl:
> > All navigation links are not grayed out when the tip is focused. At the moment I don't have a solution for this.
> 
> And right now then the solution is to put it in a section with a header, and that fixes it? @SandroDahl

No this has nothing to do with the header. The header was simply intended as a beautification.

 I just noticed the problem today. On the screenshot you can see that a link is grayed out automatically when the tip is displayed/focused. It doesn't work automatically for a navigation link or if a link has an asset icon. I think this problem should be solved in a follow up bug issue.

mikaelacaron:
> I think this problem should be solved in a follow up bug issue.

I agree, go ahead and address the other commends I had for now, and I'll make a new issue for this bug after this PR is merged!

SandroDahl:
> > I think this problem should be solved in a follow up bug issue.
> 
> I agree, go ahead and address the other commends I had for now, and I'll make a new issue for this bug after this PR is merged!

I think everything is finished. Please re-review.

mikaelacaron (approved):
Great work!! Thanks!

mikaelacaron:
Also note that the PR description should've said `Closes #214`
Then it would've automatically closed the PR when I merged the PR

PR #232, opened 2023-10-27
Updating an Odometer Reading

### LINKED ISSUE #211: Updating an Odometer Reading
* Add a swipe action to swipe left on a row for an odometer reading on the `OdometerView` and show an edit view as a sheet
* The sheet should be called `EditOdometerReadingView` and put in the correct folder
* Show a sheet to edit the values for that `OdometerReading`

See how the code works for editing a `MaintenanceEvent` on the `DashboardView`, this view should work in the same way

* N/A

ShaniDan:
# What it Does
* Closes #211 

1.  OdometerView 
-  SwipeAction to swipe left on a row for an odometer reading and show an edit view as a sheet
2.  EditOdometerReadingView
-  View that updates the Odometer Reading and is shown as a sheet in the OdometerView
3. OdometerViewModel
-   updateReading function that updates the data in firebase

# How I Tested
* Run the application to make sure that there aren't any errors
* Previewed EditOdometerReadingView in Xcode preview to see the Update Odometer view but wasn't able to test if it's working correctly due to not having credentials to Firebase Firestore

# Notes
* N/A

# Screenshot

Screenshot 2023-10-26 at 7 44 53 PM

### DIFF
--- a/Basic-Car-Maintenance/Shared/Odometer/Views/EditOdometerReadingView.swift
+++ b/Basic-Car-Maintenance/Shared/Odometer/Views/EditOdometerReadingView.swift
@@ -1,0 +1,101 @@
+//
+//  EditOdometerReadingView.swift
+//  Basic-Car-Maintenance
+//
+//  Created by Shakhnoza Mirabzalova on 10/26/23.
+//
+
+import FirebaseAnalyticsSwift
+import SwiftUI
+
+struct EditOdometerReadingView: View {
+    
+    @Binding var selectedOdometerReading: OdometerReading?
+    var viewModel: OdometerViewModel
+    @State private var date = Date()
+    @State private var selectedVehicle: Vehicle?
+    @State private var isMetric = false
+    @State private var distance = 0
+    @Environment(\.dismiss) var dismiss
+    
+    var body: some View {
+        NavigationStack {
+            Form {
+                Section {
+                    HStack {
+                        TextField("Distance", value: $distance, format: .number)
+                        
+                        Picker(selection: $isMetric) {
+                            Text("Miles").tag(false)
+                            Text("Kilometers").tag(true)
+                        } label: {
+                            Text("Preferred units", comment: "Label for units selected when adding an odometer reading")
+                        }
+                        .pickerStyle(.segmented)
+                    }
+                }
+                        Section {
+                            Picker(selection: $selectedVehicle) {
+                                ForEach(viewModel.vehicles) { vehicle in
+                                    Text(vehicle.name).tag(vehicle as Vehicle)
+                                }
+                            } label: {
+                                Text("Select a vehicle", comment: "Picker for selecting a vehicle")
+                            }
+                            .pickerStyle(.menu)
+                        } header: {
+                            Text("Vehicle",
+                                 comment: "Odometer reading vehicle picker header")
+                        }
+                        DatePicker(selection: $date, displayedComponents: .date) {
+                            Text("Date")
+                        }
+            }
+            .analyticsScreen(name: "\(Self.self)")
+            .onAppear {
+                // MARK: need to add appear
+                guard let selectedOdometerReading = selectedOdometerReading else { return }
+                setOdometerReadingValues(reading: selectedOdometerReading)
+            }
+            .navigationTitle(Text("Update Odometer"))
+            .toolbar {
+                ToolbarItem(placement: .topBarLeading) {
+                    Button {
+                        dismiss()
+                    } label: {
+                        Text("Cancel")
+                    }
+                }
+                
+                ToolbarItem(placement: .topBarTrailing) {
+                    Button {
+                        // MARK: Need to add update action
+                        if let selectedVehicle, let selectedOdometerReading {
+                            var reading = OdometerReading(date: date, distance: distance, isMetric: isMetric, vehicle: selectedVehicle)
+                            reading.id = selectedOdometerReading.id
+                            Task {
+                                // MARK: need to add update to viewmodel
+                                await viewModel.updateReading(reading)
+                                dismiss()
+                            }
+                        }
+                    } label: {
+                        Text("Update")
+                    }
+                }
+            }
+
+        }
+    }
+    
+    func setOdometerReadingValues(reading: OdometerReading) {
+        self.distance = reading.distance
+        self.date = reading.date
+        self.isMetric = reading.isMetric
+        self.selectedVehicle = reading.vehicle
+    }
+}
+
+#Preview {
+    EditOdometerReadingView(selectedOdometerReading: .constant(OdometerReading(date: Date(), distance: 0, isMetric: false, vehicle: Vehicle(name: "", make: "", model: ""))), viewModel: OdometerViewModel(authenticationViewModel: AuthenticationViewModel()))
+}
--- a/Basic-Car-Maintenance/Shared/Odometer/Views/EditOdometerReadingView.swift
+++ b/Basic-Car-Maintenance/Shared/Odometer/Views/EditOdometerReadingView.swift
@@ -10,12 +10,15 @@ import SwiftUI
 
 struct EditOdometerReadingView: View {
     
+    let vehicles: [Vehicle]
+    
     @Binding var selectedOdometerReading: OdometerReading?
     var viewModel: OdometerViewModel
     @State private var date = Date()
     @State private var selectedVehicle: Vehicle?
     @State private var isMetric = false
     @State private var distance = 0
+    @State private var selectedVehicleID: String?
     @Environment(\.dismiss) var dismiss
     
     var body: some View {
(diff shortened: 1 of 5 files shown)


mikaelacaron (changes requested):
> Previewed `EditOdometerReadingView` in Xcode preview to see the Update Odometer view but wasn't able to test if it's working correctly due to not having credentials to Firebase Firestore

@ShaniDan You do **not** need any Firebase Firestore credentials, Please test this feature by running the application before I review it. Once you confirm it runs as expected, click the re-review button.

Where do you think you need Firebase credentials?

Re-review button

ShaniDan:
Hi Mikeala, 
I ran the app and added a new reading as shown in the attached screen recording but the reading that I added isn't being displayed in the Odometer view. The update odometer feature is shown when the list item is swiped left  but the odometer view remains empty after the new reading is added. 
Please let me know if I'm testing the application incorrectly. 
Thank you! 

https://github.com/mikaelacaron/Basic-Car-Maintenance/assets/106402127/01eaf75d-0839-458f-8ea8-e3e2f61ec5fc

mikaelacaron:
@ShaniDan  You need to select a Vehicle, this just says "select a vehicle" and doesn't show a picker because you haven't added one yet

Go to settings, add a vehicle, and then go and add a new odometer reading

ShaniDan:
@mikaelacaron Thank you for clarification. :) Below is the screen recording for the update odometer  feature.
 
https://github.com/mikaelacaron/Basic-Car-Maintenance/assets/106402127/22fbb0b1-4b55-4fe8-bee2-85fb569bb8ce

Basic-Car-Maintenance/Shared/Odometer/Views/EditOdometerReadingView.swift
```diff
@@ -0,0 +1,101 @@
...
+                            Text("Vehicle",
+                                 comment: "Odometer reading vehicle picker header")
+                        }
+                        DatePicker(selection: $date, displayedComponents: .date) {
+                            Text("Date")
+                        }
+            }
+            .analyticsScreen(name: "\(Self.self)")
+            .onAppear {
+                // MARK: need to add appear
```
mikaelacaron: Remove all the `MARK` comments you created, they're unnecessary and kind of out of place

Basic-Car-Maintenance/Shared/Odometer/Views/EditOdometerReadingView.swift
```diff
@@ -0,0 +1,101 @@
...
+                        } label: {
+                            Text("Preferred units", comment: "Label for units selected when adding an odometer reading")
+                        }
+                        .pickerStyle(.segmented)
+                    }
+                }
+                        Section {
+                            Picker(selection: $selectedVehicle) {
+                                ForEach(viewModel.vehicles) { vehicle in
+                                    Text(vehicle.name).tag(vehicle as Vehicle)
```
mikaelacaron: The indentation of this section is off. Fix it with `Ctrl` + `i`

And this picker does not work. See the `AddMaintenanceView` for the proper way to set up the picker

Basic-Car-Maintenance/Shared/Odometer/Views/OdometerView.swift
```diff
@@ -12,9 +12,11 @@ struct OdometerView: View {
     @Environment(ActionService.self) var actionService
     
     @State private var viewModel: OdometerViewModel
+    @State private var isShowingEditView = false
+    @State private var selectedOdometerReading: OdometerReading?
 
     init(authenticationViewModel: AuthenticationViewModel) {
-        viewModel = OdometerViewModel(authenticationViewModel: authenticationViewModel)
+        _viewModel = State(initialValue: OdometerViewModel(authenticationViewModel: authenticationViewModel))
```
mikaelacaron, line 18: Does this make a difference to write it this way vs the way it was written originally?

mikaelacaron (changes requested):
Great work!! I have several comments that need resolved please. If you have any questions, you can comment directly on that thread, otherwise click the resolve button once you've addressed my comment. Once you fix everything, please click the re-review button 

And note you need to update your feature branch from the latest changes on `dev`

Re-review button

Basic-Car-Maintenance/Shared/Odometer/Views/EditOdometerReadingView.swift
ShaniDan: I revised the picker as per your suggestion, and now I can select the vehicle. However, I noticed that the odometer number remains the same. My understanding is that the purpose of having the picker is to display all the values based on the selected vehicle, right?

Wouldn't it be more logical to display the name of the vehicle instead of using the picker, especially since we're editing a specific item in the odometer view? 
I believe there should be a way to update all the values based on the selected vehicle, but I couldn't figure it out.

Attached are the screen recordings for the view with the picker and with just a name of the vehicle.

https://github.com/mikaelacaron/Basic-Car-Maintenance/assets/106402127/a4d48c82-31bd-428e-a47c-89db8c7f9fa7

https://github.com/mikaelacaron/Basic-Car-Maintenance/assets/106402127/f5ab9203-ed7c-403e-b585-c44b1fd7216b

mikaelacaron: > My understanding is that the purpose of having the picker is to display all the values based on the selected vehicle, right?

No. The purpose of the picker is for some reason you wanted to change which vehicle this reading is for

> Wouldn't it be more logical to display the name of the vehicle instead of using the picker, especially since we're editing a specific item in the odometer view?

No, the goal is to change any value for this reading.

> I believe there should be a way to update all the values based on the selected vehicle, but I couldn't figure it out.

No that is not correct. Do not do this.

---

The purpose of this view is to be able to change any of the values for this specific reading. The first video you showed does not work properly because the vehicle does not change when you changed it in the picker. The 2nd video is not the correct behavior, because I **do** want the user to be able to change the vehicle.

Basic-Car-Maintenance/Shared/Odometer/Views/OdometerView.swift
ShaniDan: It gave me an error "Variable 'self.viewModel' used before being initialized" when I added `@State private var selectedOdometerReading: OdometerReading` and I used `_viewModel = State(initialValue:` inside the initialization to resolve the error. 
Please let me know if you think if it's not the correct way. 
Thank you!

mikaelacaron: That's because you need to give the `selectedOdometerReading` a value

Only defining `@State private var selectedOdometerReading: OdometerReading` does not set a value, and will result in that warning

You need to make `selectedOdometerReading` optional (which you did) and that warning would've have been there. but this initializer doesn't need to change if `selectedOdometerReading` is optional

 Another way is to pass it as an input variable, and then you'd still need something to hold all the different text fields as state properties

mikaelacaron (changes requested):
@ShaniDan I'll accept this for Hacktoberfest, but you haven't addressed any of the comments for merging this PR, before clicking the re-review button. Please address those, and then click the review button

Basic-Car-Maintenance/Shared/Odometer/Views/EditOdometerReadingView.swift
ShaniDan: I've set the vehicle.id as the tag for the Picker(selection: $selectedVehicleID) to ensure that the selection works for the vehicles in the picker.
```
Picker(selection: $selectedVehicleID) {
              Text(vehicle.name).tag(vehicle.id)
            }
```
However, I encountered a challenge in updating the odometer reading based on the selected vehicle in the picker. The editOdometerReadingView is effectively updating the odometer for the vehicle listed in the OdometerView.

It seems that the issue might be related to the use of `@Binding var selectedOdometerReading: OdometerReading`. This variable is utilized in both the onAppear and the Button responsible for updating the odometer reading.

I've tried alternative approaches without the use of `@Binding var selectedOdometerReading: OdometerReading`, but I encountered some errors. I'd like to find the solution for the issue as well. As a newcomer to iOS development, I might've overlooked certain aspects, and I appreciate your understanding and patience.

Basic-Car-Maintenance/Shared/Odometer/Views/OdometerView.swift
ShaniDan: I gave a value to the selectedOdometerReading  `@State private var selectedOdometerReading = OdometerReading( date: Date(), distance: 100, isMetric: true, vehicle: Vehicle(name: "Example Vehicle", make: "Make", model: "Model"))` but it gave me an error "Cannot convert value of type 'Binding<OdometerReading>' to expected argument type 'Binding<OdometerReading?>'"  on the
 
`.sheet(isPresented:` $isShowingEditView) {
                        EditOdometerReadingView(vehicles: viewModel.vehicles, selectedOdometerReading: $selectedOdometerReading, viewModel: viewModel) }`

Basic-Car-Maintenance/Shared/Odometer/Views/EditOdometerReadingView.swift
mikaelacaron: > I've set the `vehicle.id` as the tag for the `Picker(selection: $selectedVehicleID)` to ensure that the selection works for the vehicles in the picker

Can you commit this and push the changes? So I can see what you're doing on GitHub which can help me point you in the right direction

> but I encountered some errors

Can you elaborate more what the error were and what line

Look at the `EditMaintenanceEventView` and the `DashboardView` for how this works, you're doing essentially the same thing, but with odometer readings

ShaniDan: Hi @mikaelacaron
I pushed the commit from my local branch to dev. Can you please check if you can see it? 
image

The picker in the `EditMaintenanceEventView `isn't working properly to select the vehicles.

mikaelacaron:
Closing now that Hacktoberfest is over, and I don't want to merge this solution right now. Thanks again for contributing! It's much appreciated! Feel free to pick up any issues even though Hacktoberfest is over

PR #238, opened 2023-10-28
Fix `DashboardView` no events on 1st load

### LINKED ISSUE #210: BUG - `DashboardView` events Aren't Loading on the First Load
## Describe the bug
* The `MaintenanceEvent`s on the `DashboardView` aren't loading, when the app is first loaded. Instead it's showing the empty screen that says "Add your first maintenance event"

## To Reproduce
1.  Add at least one vehicle and at least one maintenance event
2. Make sure the first tab that is going to open is the `Dashboard` (it saves the last tab when the app was closed)
2. Run the app and wait for the app to load and show events
3. The events never load, and it only shows the "empty view" that says `Add your first maintenance`

## Expected behavior
* The `Dashboard` tab loads the data even when it's the first tab to be open, you do **NOT** need to tap to a different tab, and tap back to see data

## Screenshots
* N/A

 * OS: iOS 17.0
 * App Version: v1.0 (1)

## Additional Context
* Add any other context about the problem here.

hanhqvu:
# What it Does
* Closes #210 
* This PR will call the closure to get the events & vehicles when the app is in the foreground, not when the view appear

# How I Tested
* Build then run the app in the simulator

# Notes
* I'm not entirely sure if this is best practice

# Screenshot
RocketSim_Recording_iPhone_15_2023-10-28_22 18 34

### DIFF
--- a/Basic-Car-Maintenance/Shared/Dashboard/Views/DashboardView.swift
+++ b/Basic-Car-Maintenance/Shared/Dashboard/Views/DashboardView.swift
@@ -114,16 +114,17 @@ struct DashboardView: View {
                     }
                 }
             }
-            .task {
-                await viewModel.getMaintenanceEvents()
-                await viewModel.getVehicles()
-            }
             .sheet(isPresented: $isShowingAddView) {
                 makeAddMaintenanceView()
             }
         }
         .onChange(of: scenePhase) { _, newScenePhase in
             guard case .active = newScenePhase else { return }
+            
+            Task {
+                await viewModel.getMaintenanceEvents()
+                await viewModel.getVehicles()
+            }
             
             guard let action = actionService.action,
                   action == .newMaintenance
--- a/Basic-Car-Maintenance/Shared/Dashboard/Views/DashboardView.swift
+++ b/Basic-Car-Maintenance/Shared/Dashboard/Views/DashboardView.swift
@@ -146,9 +146,6 @@ struct DashboardView: View {
     private func makeAddMaintenanceView() -> some View {
         AddMaintenanceView(vehicles: viewModel.vehicles) { event in
             viewModel.addEvent(event)
-            Task {
-                await viewModel.getMaintenanceEvents()
-            }
         }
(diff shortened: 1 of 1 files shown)


mikaelacaron (changes requested):
@hanhqvu No I would not solve it this way, is this not being called properly in the `.task`?

Also the code you removed from the `makeAddMaintenanceView` function will cause a bug, why is this removed?

hanhqvu:
@mikaelacaron I added breakpoints to `.task` and it is not called when `DashboardView` appear on 1st load. If switch tabs then it got called. Regarding the removed code, I think the bug is tied to `.task` not being called so not getting vehicles.

mikaelacaron:
> Regarding the removed code, I think the bug is tied to .task not being called so not getting vehicles.

No the point of line 149 to 151

is to recall the events again after an event has been added, to update the list

> I added breakpoints to .task and it is not called when DashboardView appear on 1st load.

I totally thought it should be, could you find documentation behind this why it's not called when the view first appears?

hanhqvu:
> > Regarding the removed code, I think the bug is tied to .task not being called so not getting vehicles.
> 
> 
> 
> No the point of line 149 to 151
> 
> 
> 
> is to recall the events again after an event has been added, to update the list
> 
> 
> 
> > I added breakpoints to .task and it is not called when DashboardView appear on 1st load.
> 
> 
> 
> I totally thought it should be, could you find documentation behind this why it's not called when the view first appears?

Ah, I see. My bad on that, I'll revert that.

Also another point, could be because I'm not familiar with `Firebase` but why does the events get cleared out of memory when the app goes into background. The events should still be in memory if the app is still in memory right?

mikaelacaron:
> could be because I'm not familiar with Firebase but why does the events get cleared out of memory when the app goes into background

This.... I'm not sure.... what scenario did you notice this behavior?

hanhqvu:
Prior to the my changes after events have been fetched, when I go to the home screen then re-open the app, the events are no longer there.

PR #247, opened 2023-10-30
implement Add Odometer Reading AppIntent

### LINKED ISSUE #222: FEATURE - Make an App Intent for Adding an Odometer Reading with Shortcuts
* Use App Intents to create a shortcut to add a new `OdometerReading`

* N/A

## Additional context
* Here's one of the App Intent WWDC [videos](https://developer.apple.com/wwdc22/10170) to help get started
* This is part of a bigger idea with automatic odometer readings, see [this discussion](https://github.com/mikaelacaron/Basic-Car-Maintenance/discussions/145#discussion-5723833), but don't do this whole feature, with the automatic logging, this issue is only for making a shortcut with App Intents
* I haven't used this yet, but if any entitlement things or targets need to be added, let me know and I can add that, and then you can continue on this (rather than you adding anything), because of signing issues and merge conflicts that can happen

OmarHegazy93:
# What it Does
* Closes #222
Implement shortcut for adding new odometer reading

### To achieve this task, I had to make some refactoring for the app flow by:
1. Removing the dependency for the `AuthenticationViewModel` for almost all viewModels, as they all needed the UserID only 
2. UserID is currently saved to `UserDefaults`, as a next step we will need to figure out a better or _safer_ way for saving it
3. Make `getVehicles()` method default implemented as both `MainTabViewModel` and `SettingsViewModel` are using it for fetching vehicles

### For the AppIntent itself:
1. The query used to fetch the vehicles is using `MainTabViewModel` and if no vehicles available, then it will return an error
2. `DistanceUnit` enum is used to make more readable for selecting Distance unit
3. Validation is applied to `Distance` and `Date`

# How I Tested
###  Happy scenario
1. Install the app
2. Make sure at least one vehicle is added in the app by going to `Settings` tab > `Add Vehicle`
3. Switch to Shortcuts app
4. Tab on the `+` on the top right corner
5. Tab on `⨁ Add Action`
6. Select `Apps` tab
7. Tab on `Basic Car`
8. Tab on `Add Odometer Reading`
9. Add data for all fields and tab the play button on the bottom right corner
10. See the result

###  Unhappy scenario
1. Install the app
2. Make sure at least one vehicle is added in the app by going to `Settings` tab > `Add Vehicle`
3. Switch to Shortcuts app
4. Tab on the `+` on the top right corner
5. Tab on `⨁ Add Action`
6. Select `Apps` tab
7. Tab on `Basic Car`
8. Tab on `Add Odometer Reading`
9. Try to set distance value to zero or negative value or set the date to past
10. See the result

# Notes

# Screenshot

### DIFF
--- a/Basic-Car-Maintenance/Shared/Dashboard/ViewModels/DashboardViewModel.swift
+++ b/Basic-Car-Maintenance/Shared/Dashboard/ViewModels/DashboardViewModel.swift
@@ -10,17 +10,13 @@ import FirebaseFirestoreSwift
 import Foundation
 
 @Observable
-class DashboardViewModel {
-    
-    let authenticationViewModel: AuthenticationViewModel
-    
+class DashboardViewModel: ObservableObject {
     var events = [MaintenanceEvent]()
     var showAddErrorAlert = false
     var showErrorAlert = false
     var isShowingAddMaintenanceEvent = false
     var errorMessage: String = ""
     var sortOption: SortOption = .custom
-    var vehicles = [Vehicle]()
     var searchText: String = ""
     var isLoading = false
 
@@ -40,12 +36,8 @@ class DashboardViewModel {
         }
     }
     
-    init(authenticationViewModel: AuthenticationViewModel) {
-        self.authenticationViewModel = authenticationViewModel
-    }
-    
     func addEvent(_ maintenanceEvent: MaintenanceEvent) {
-        if let uid = authenticationViewModel.user?.uid {
+        if let uid = AppDefaults.getUserID() {
             var eventToAdd = maintenanceEvent
             eventToAdd.userID = uid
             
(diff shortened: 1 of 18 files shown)


Basic-Car-Maintenance/Shared/Dashboard/ViewModels/DashboardViewModel.swift
```diff
@@ -10,17 +10,13 @@ import FirebaseFirestoreSwift
 import Foundation
 
 @Observable
-class DashboardViewModel {
-    
-    let authenticationViewModel: AuthenticationViewModel
-    
+class DashboardViewModel: ObservableObject {
```
mikaelacaron, line 13: I don't think this should be conforming to `ObservableObject`, and using the `@Observable` macro

mikaelacaron (commented):
I don't want to introduce this big of a rewrite at this moment, but thanks for all the work on this feature!
I'll revisit it after Hacktoberfest

mikaelacaron:
Closing now that Hacktoberfest is over, and I don't want to merge this solution right now. Thanks again for contributing! It's much appreciated! Feel free to pick up any issues even though Hacktoberfest is over

PR #248, opened 2023-10-30
Change Label Icons on SettingsView to use Scalable Images

### LINKED ISSUE #103: IMPROVE - Scale Images on `SettingsView`
* When the dynamic type size of the `SettingsView` is at the largest size, you can see how the images don't scale (only the Contributors row does)
* Scale the icons as the dynamic type scales

IMG_5493

darrengillman:
# What it Does
* Closes #103
* Scales the images in Label icons within the SettingsView in parallel with changes in Dynamic Text
* Uses an @ScaledMetric property wrapper, based on the .largeTitle font, to define the image frame dimensions 

# How I Tested
* Ran the original and modified code in parallel in two simulators
* changed the dynamic type size in both simulators using cmd-opt-plus & cmd-opt-minus
* visually confirmed that the modified code scaled the images up and down in line with changes to dynamic text

# Notes
Scaling the text to higher font sizes highlights the layout discrepancy between the Labels for most items in the List and the use of a single Text view for the maintainer row.  Consider changes this to a Label too?

# Screenshot
BMC-103

### DIFF
--- a/Basic-Car-Maintenance/Shared/Settings/Views/SettingsView.swift
+++ b/Basic-Car-Maintenance/Shared/Settings/Views/SettingsView.swift
@@ -14,6 +14,8 @@ struct SettingsView: View {
     @Environment(ActionService.self) var actionService
     @Environment(\.scenePhase) var scenePhase
     @Environment(\.colorScheme) var colorScheme
+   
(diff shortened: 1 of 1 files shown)


mikaelacaron (approved):
Great work!! Thanks for contributing!

PR #236, opened 2023-10-27
feature: update german localization

appfrosch:
# What it Does
* Closes #233
* Updates the German localization

PR #251, opened 2023-10-30
update the german localization

### LINKED ISSUE #233: IMPROVE - German Localization
## Localizations: update German translation
The German translation is outdated and needs some work.

n/a

appfrosch:
# What it Does
* Closes #233
- updates the german localization to be 100% localized

### DIFF
--- a/Basic-Car-Maintenance/Shared/Localizable.xcstrings
+++ b/Basic-Car-Maintenance/Shared/Localizable.xcstrings
@@ -18,7 +18,7 @@
         "de" : {
           "stringUnit" : {
             "state" : "translated",
-            "value" : ""
+            "value" : "%@"
           }
         },
         "es-419" : {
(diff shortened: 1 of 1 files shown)


appfrosch:
This is the second attempt to clear that issue, the first pr #236 was messed up due to merge conflicts with dev.

Basic-Car-Maintenance/Shared/Localizable.xcstrings
```diff
@@ -61,6 +61,12 @@
     },
     "%lld %@" : {
       "localizations" : {
+        "de" : {
+          "stringUnit" : {
+            "state" : "translated",
+            "value" : ""
+          }
+        },
```
mikaelacaron: I think this shouldn't be empty, but should be `%lld %@` because this is filled in automatically, from computed properties and such, if it was empty, I think it would end up showing as an empty string, when there should be a value there

mikaelacaron (approved):
Great work! Thanks for the update! I'll be pushing one commit related to my comment

appfrosch:
That makes a whole lot of sense, I was assuming that it would take the default English value in this case, so TIL.

Thanks for the additional commit and letting me know.

PR #252, opened 2023-10-30
Added Turkish Localization

isiko35:
# What it Does
* #7 
* Turkish localization

# How I Tested
* Set device language to Turkish
* Check screens for visual confirmation

# Notes
* It will be an ongoing effort to check for new strings to translate during development

# Screenshot
Simulator Screenshot - iPhone 15 Pro - 2023-10-30 at 15 29 40

### DIFF
--- a/Basic-Car-Maintenance/Shared/Localizable.xcstrings
+++ b/Basic-Car-Maintenance/Shared/Localizable.xcstrings
@@ -51,6 +51,12 @@
             "value" : "%@"
           }
         },
+        "tr" : {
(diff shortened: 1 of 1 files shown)


mikaelacaron (approved):
Thanks for adding Turkish!

mikaelacaron:
Also note a better branch name would've been `7-add-turkish-localization`, without the `#` because that's not a best practice in git branch names

isiko35:
Thanks @mikaelacaron, I will do that next time.

PR #253, opened 2023-10-30
IMPROVE - Analytics Screen View Modifier #197

### LINKED ISSUE #197: IMPROVE - Analytics Screen View Modifier
* Right now there are 8 views that have this modifier`.analyticsScreen(name: "\(Self.self)")` on them
* Create a new file called `FirebaseAnalytics+Extension` with a new view modifier that uses this modifier (put this file in the `Utilities` folder)
   * The new modifier should be called `analyticsView`
* Go through and use the newly created modifier on each view and remove the `import FirebaseAnalyticsSwift` on all the views, where necessary

The goal is that the new view modifier file has only a single import, and that we no longer need to import `FirebaseAnalyticsSwift` on all the views 

* N/A

jessielinden:
# What it Does
* Closes #197 
* Created new file FirebaseAnalytics+Extension and removed all other imports of FirebaseAnalyticsSwift

# How I Tested
* Upon launch, I see many Analytics printouts in the console, so while I am not at all familiar with Firebase I think the modifier was successful.


mikaelacaron (approved):
Great work!! Yeah technically you'd only see logs in the console, only I can see the actual logs set in Firebase

PR #255, opened 2023-10-31
Add Dutch Localization

### LINKED ISSUE #7: FEATURE - Localization (Multiple assignees allowed!)

krz:
# adds Dutch language localization
* Closes #7 
* translation for all fields


mikaelacaron (approved):
Great work!! Note that you should've made a branch and committed there, rather than directly committing to your `dev` branch, because I think from now on, it'll show a commit on there, but it technically shouldn't (You may need to re-fork, the repo)

PR #256, opened 2023-10-31
Firebase Local Emulator

### LINKED ISSUE #14: Setup the Firebase Local Emulator
## Details
* Following [this tutorial](https://peterfriese.github.io/MakeItSo/tutorials/makeitso/02-developing-locally-with-the-emulator-suite), set up the project to use the Firebase local emulator when developing and use real Firebase for production
* Update the README.md and CONTRIBUTING.md with instructions about how to use the Firebase local emulator, if needed (if you're not sure what to say for set up, comment in the PR and Mikaela can update it)

* N/A

## Additional context
* N/A

n0rthk1n9:
# What it Does
* Closes #14
* It makes the app testable against a local emulated Firebase Auth and Firebase Firestore instance

# How I Tested
* Make sure Node v 8.0 or higher is installed
* Make sure Openjdk v 11 or higher is installed (If you installed it via brew, don't forget to add the executable to $PATH, brew will output it as a handy command to execute for you)
* Make sure firebase-cli is installed (`npm install -g firebase-tools`)
* Make sure you ran  `firebase login` and authenticated successfully
* Go to `backend/Basic-Car-Maintenance` and run `firebase emulators:start`
* Now every request from the app should go against the emulator
* ﻿﻿In the emulator UI at http://127.0.0.1:4000/ you can check if the data you added in the app is added in the emulated Firestore

# Notes

# Screenshot
SCR-20231031-mvu


Basic-Car-Maintenance.xcodeproj/xcshareddata/xcschemes/Basic-Car-Maintenance.xcscheme
```diff
@@ -73,6 +73,12 @@
             ReferencedContainer = "container:Basic-Car-Maintenance.xcodeproj">
          </BuildableReference>
       </BuildableProductRunnable>
+      <CommandLineArguments>
+         <CommandLineArgument
+            argument = "-useEmulator YES"
+            isEnabled = "YES">
+         </CommandLineArgument>
+      </CommandLineArguments>
```
mikaelacaron, line 81: When would this flag need to be enabled vs disabled? Only when wanting to use the emulator? Otherwise it would be unchecked, which would default to using the real project?

Basic-Car-Maintenance/Shared/BasicCarMaintenanceApp.swift
```diff
@@ -34,6 +36,17 @@ class AppDelegate: NSObject, UIApplicationDelegate {
         didFinishLaunchingWithOptions launchOptions: [UIApplication.LaunchOptionsKey: Any]? = nil
     ) -> Bool {
         FirebaseApp.configure()
+        
+        let useEmulator = UserDefaults.standard.bool(forKey: "useEmulator")
+        if useEmulator {
```
mikaelacaron, line 41: Why use UseDefaults, as opposed to the command line argument?

mikaelacaron (changes requested):
Great work!! I haven't tested this because I haven't downloaded everything

> * Make sure Node v 8.0 or higher is installed
> * Make sure Openjdk v 11 or higher is installed (If you installed it via brew, don't forget to add the executable to $PATH, brew will output it as a handy command to execute for you)
> * Make sure firebase-cli is installed (npm install -g firebase-tools)


Can you note the best way to download these? (Preferably with Homebrew) Also I have a few questions, you can comment directly in that thread

Basic-Car-Maintenance.xcodeproj/xcshareddata/xcschemes/Basic-Car-Maintenance.xcscheme
n0rthk1n9: Yes exactly

n0rthk1n9:
The best way to install all the requirements is indeed via homebrew:

1. `brew install nvm` (Installs node version manager, so you don't need to update the system node version)
2. `nvm install stable` and `nvm use stable` to download and set the last stable node version
3. `brew install openjdk` to install openjdk. Add the executable to the $PATH via .zshrc or .bashrc file as prompted after installation
4. npm install -g firebase-tools

With that all prerequisites should be installed. Let me know if anything doesn't work

Basic-Car-Maintenance/Shared/BasicCarMaintenanceApp.swift
n0rthk1n9: Honestly I don't have a better answer to that question than "Because Peter recommended it doing it like that in his article" But I guess it could also be done via command line arguments, if you prefer that :)

mikaelacaron:
Thanks so much for all the help on this! I'm finally getting around to implementing it

I'm going to close your PR because I'm working on this on a separate branch, because I'm also updating a few other things.

Your steps were super helpful to finally get this running!

PR #264, opened 2023-11-17
POC of generating screenshots with a Unit Test

Dean151:
It uses my personal library that I open sourced just now. Still early in the work, but improving… Screenshots are generated in a folder aside UT, and can be moved around in a fastlane script that execute those tests.

Capture d’écran 2023-11-17 à 17 25 30

### DIFF
--- a/Basic-Car-Maintenance-ScreenshotsTests/ScreenshotsTests.swift
+++ b/Basic-Car-Maintenance-ScreenshotsTests/ScreenshotsTests.swift
@@ -1,0 +1,25 @@
+//
+//  ScreenshotsTests.swift
+//  Basic-Car-Maintenance-ScreenshotsTests
+//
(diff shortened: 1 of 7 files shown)


mikaelacaron:
This is cool!! I think I'm not going to implement it just yet, because I don't want to use a package for this, at the moment, but it's really cool!

Should I make a "test plan" for unit tests and then there would be a separate one for UI tests? @Dean151

Dean151:
> This is cool!! I think I'm not going to implement it just yet, because I don't want to use a package for this, at the moment, but it's really cool!
> 
> Should I make a "test plan" for unit tests and then there would be a separate one for UI tests? @Dean151

My implementation is more like:
- UT & UI tests targets for testing the app
- Separated UT for screenshots, that we run when needed

PR #269, opened 2023-11-21
5 Edit Vehicle

### LINKED ISSUE #5: FEATURE - Edit Vehicle Details

TratonGossink:
# What it Does
* Closes #5 Edit vehicle 
* This change allows the user to edit an existing vehicle.

# How I Tested
*Run the application.
*Tap "Add Vehicle"
*Create new mock-up vehicle
*Swip to view edit option

# Notes
* N/A

# Screenshot

### DIFF
--- a/Basic-Car-Maintenance/Shared/Localizable.xcstrings
+++ b/Basic-Car-Maintenance/Shared/Localizable.xcstrings
@@ -4280,6 +4280,7 @@
       }
     },
     "Update" : {
+      "comment" : "Label to update vehicle information.",
       "localizations" : {
         "be" : {
           "stringUnit" : {
@@ -4407,6 +4408,9 @@
           }
         }
       }
+    },
+    "Update Vehicle" : {
+
     },
     "Vehicle" : {
       "comment" : "Maintenance event vehicle picker header",
--- a/Basic-Car-Maintenance/Shared/Localizable.xcstrings
+++ b/Basic-Car-Maintenance/Shared/Localizable.xcstrings
@@ -1518,7 +1518,7 @@
       }
     },
     "Edit" : {
-      "comment" : "Button label to edit this maintenance\nButton label to edit this vehicle",
+      "comment" : "Button label to edit this vehicle\nButton label to edit this maintenance",
       "localizations" : {
         "be" : {
           "stringUnit" : {
@@ -4409,7 +4409,7 @@
         }
       }
     },
-    "Update Vehicle" : {
+    "Update Vehicle Info" : {
 
     },
     "Vehicle" : {
--- a/Basic-Car-Maintenance/Shared/Localizable.xcstrings
+++ b/Basic-Car-Maintenance/Shared/Localizable.xcstrings
@@ -1518,7 +1518,7 @@
       }
     },
     "Edit" : {
-      "comment" : "Button label to edit this vehicle\nButton label to edit this maintenance",
+      "comment" : "Button label to edit this maintenance\nButton label to edit this vehicle",
       "localizations" : {
         "be" : {
           "stringUnit" : {
--- a/Basic-Car-Maintenance/Shared/Localizable.xcstrings
+++ b/Basic-Car-Maintenance/Shared/Localizable.xcstrings
@@ -1518,7 +1518,7 @@
       }
     },
     "Edit" : {
-      "comment" : "Button label to edit this maintenance",
+      "comment" : "Button label to edit this vehicle\nButton label to edit this maintenance",
       "localizations" : {
         "be" : {
           "stringUnit" : {
--- a/Basic-Car-Maintenance/Shared/Settings/ViewModels/SettingsViewModel.swift
+++ b/Basic-Car-Maintenance/Shared/Settings/ViewModels/SettingsViewModel.swift
@@ -17,6 +17,8 @@ final class SettingsViewModel {
     var contributors: [Contributor]?
     
     var vehicles = [Vehicle]()
+    var errorMessage: String = ""
+    var showErrorAlert = false
     
     var sortedContributors: [Contributor] {
         guard let contributors = contributors, !contributors.isEmpty else {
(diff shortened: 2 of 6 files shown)


Basic-Car-Maintenance/Shared/Localizable.xcstrings
```diff
@@ -1518,7 +1518,7 @@
       }
     },
     "Edit" : {
-      "comment" : "Button label to edit this maintenance",
+      "comment" : "Button label to edit this vehicle\nButton label to edit this maintenance",
```
mikaelacaron, line 1521: This is a thing I recently learned, you can't have two comments for a single key
So for this you'd need to make a new key like `EditVehicle` and then you'd write the commend, and in the actual Localizable.xcstrings file you'd write just `Edit`

Basic-Car-Maintenance/Shared/Settings/ViewModels/SettingsViewModel.swift
```diff
@@ -88,6 +90,30 @@ final class SettingsViewModel {
...
+                    .collection(FirestoreCollection.vehicles)
+                    .document(id)
+                    .setData(from: eventToUpdate)
+            } catch {
+                showErrorAlert.toggle()
+                errorMessage = error.localizedDescription
+            }
+        }
+        
+        AnalyticsService.shared.logEvent(.maintenanceUpdate)
```
mikaelacaron, line 112: This is not a `maintenanceUpdate` this is a `vehicleUpdate`

Basic-Car-Maintenance/Shared/Settings/ViewModels/SettingsViewModel.swift
```diff
@@ -88,6 +90,30 @@ final class SettingsViewModel {
         }
     }
     
+    ///Updates users vehicle that is being edited. Fetches selected vehicle from Firestore and saves to that selected vehicle.
+    func updateEvent(_ editVehicleEvent: EditVehicleEvent) async {
```
mikaelacaron, line 94: You shouldn't be making a new model `EditVehicleEvent`, you should just be using `Vehicle`

Basic-Car-Maintenance/Shared/Settings/ViewModels/SettingsViewModel.swift
```diff
@@ -88,6 +90,30 @@ final class SettingsViewModel {
...
+                    .setData(from: eventToUpdate)
+            } catch {
+                showErrorAlert.toggle()
+                errorMessage = error.localizedDescription
+            }
+        }
+        
+        AnalyticsService.shared.logEvent(.maintenanceUpdate)
+        
+        await self.getVehicles()
```
mikaelacaron, line 114: You don't need the `self` it's implied

Basic-Car-Maintenance/Shared/Settings/Views/EditVehicleView.swift
```diff
@@ -0,0 +1,124 @@
...
+                            event.id = selectedEvent.id
+                            Task {
+                                await viewModel.updateEvent(event)
+                                dismiss()
+                            }
+                        }
+                    } label: {
+                        Text("Update")
+                    }
+                    .disabled(name.isEmpty)
```
mikaelacaron, line 100: Follow the same rules as `isVehicleValid` in the `AddVehicleView`

Basic-Car-Maintenance/Shared/Settings/Views/SettingsView.swift
```diff
@@ -23,6 +23,11 @@ struct SettingsView: View {
     @State private var errorDetails: Error?
     @State private var copiedAppVersion: Bool = false
     
+    @State private var selectedVehicleEvent: EditVehicleEvent?
+    @State private var isEditingVehicle = false
+    @State var editViewModel: EditVehicleView?
+    @State private var vehicleToEdit: Vehicle?
```
mikaelacaron, line 29: You only need the `vehicleToEdit`, you shouldn't be using a separate model for editing, so `EditVehicleEvent` is not needed

Also don't call this `editViewModel` is not needed

Basic-Car-Maintenance/Shared/Settings/Views/SettingsView.swift
```diff
@@ -124,9 +129,21 @@ struct SettingsView: View {
                             } label: {
                                 Text("Delete", comment: "Label to delete a vehicle")
                             }
+                            Button {
+                                isEditingVehicle = true
```
mikaelacaron, line 133: Here is where you would put `vehicleToEdit = vehicle`, because this is in the `ForEach` loop, and it's specifying which vehicle is going to show in the `EditVehicleView`

Basic-Car-Maintenance/Shared/Settings/Views/SettingsView.swift
```diff
@@ -124,9 +129,21 @@ struct SettingsView: View {
...
+                            } label: {
+                                VStack {
+                                    Text("Edit", comment: "Button label to edit this vehicle")
+                                    Image(systemName: SFSymbol.pencil)
+                                }
+                            }
+                        }
+                        .sheet(isPresented: $isEditingVehicle) {
+                                            EditVehicleView(
+                        selectedEvent: $selectedVehicleEvent, viewModel: viewModel)
```
mikaelacaron, line 143: Here we'd pass the `vehicleToEdit` because we are using the Vehicle model, not the new one you created

mikaelacaron (changes requested):
You're almost there!! I have several comments that need resolved, which you know from our DMs. If you have any questions, you can comment directly on that thread, otherwise click the resolve button once you've addressed my comment. Once you fix everything, please click the re-review button 

Re-review button

mikaelacaron (approved):
Hey @TratonGossink I'm going to go ahead and merge this into `fix-vehicle-event-connection` which is something I'm already working on, because I'm redoing the current Firestore schema

Thanks for all your work!

Issue #40, opened 2023-10-02
FEATURE - Onboarding UI


clydehenry3:
## Creating a 2-3 page onboarding experience for the user.
* These screens would present the highlighted features for users to capture before entering the app. 
* They would have text and visuals to ensure the user was informed of what to expect and to bring excitement before entering the app. 

## Alternatives
* Other alternatives would include coach marks after sign-up/login, which could hinder the User's Experience. 
* No onboarding/coach marks, but it would increase the likelihood of not using the app. 

## Additional context
* It Would be beneficial to create in Figma and integrate it into VS Code for a seamless integration 
* Figma now has developer options where I can export near to-ready/ready code.

clydehenry3:
I would love to tackle this issue.

horacio3m:
I would love work with u guys on this issue ![image](https://github.com/mikaelacaron/Basic-Car-Maintenance/assets/8496232/9b0e9248-5380-465b-b0c9-24ea895b9ba7)

mikaelacaron:
@horaciomota Can you DM me your email on any social media, I'm going to start a Figma board for this! (I've been DM'ing with @clydehenry3 about this too, but I haven't set it up yet), I can do this today

horacio3m:
@mikaelacaron ill DM u on instagram. Right Now. @horaciomota my iG

horacio3m:
DM sent

clydehenry3:
Awesome to have more people on board! @horaciomota @mikaelacaron

clydehenry3:
Hey @horaciomota! I would love to connect on socials so we can talk. Thanks for popping in early on Figma

horacio3m:
My username is horacio mota on all social platforms. You can reach out to me on Instagram at @horaciomota. @clydehenry3

clydehenry3:
@horaciomota Sent.

justinseal:
I was was just thinking about this. A little Onboarding to help with why a user would want to sign in and what help it might provide.

mikaelacaron:
> why a user would want to sign in
It uses Firebase Anonymous Auth! Users don't technically need to sign in unless they want to see their data on multiple devices

justinseal:
> > why a user would want to sign in
> > It uses Firebase Anonymous Auth! Users don't technically need to sign in unless they want to see their data on multiple devices

Got it, but there's nothing there that says they *can* do that. Yet.

clydehenry3:
[...]

PR #276, opened 2023-11-30
fix: Prevent Users From Deleting All Vehicles

### LINKED ISSUE #202: Prevent Users From Deleting All Vehicles
* The user shouldn't be able to delete all of their vehicles
* When they swipe to delete, once they're attempting to delete the last one, show an alert
   * **Title**: `Can't Delete Last Vehicle`
   * **Message**: `The last vehicle can't be deleted`

* N/A

## Additional context
* N/A

windrunner21:
# What it Does
* Closes #202
* I have added control to check the number of active vehicles user has, and if user has only 1 vehicle left, they cannot delete the last vehicle. If they press delete or swipe they are greeted with an alert.

# How I Tested
* Run the application, open Settings tab
* Deleted vehicle using swipe gesture.
* Was greeted with an Alert.
* Deleted vehicle pressing on delete swipe action button.
* Was greeted with an Alert.

# Notes
* Delete functionality itself wasn't implemented yet. I had already 2 vehicles added. Because I couldn't delete them from Firebase, I upped the control counter to 2 for local tests. I pushed the code with control counter set to 1.

# Screenshot
* Adding scree

https://github.com/mikaelacaron/Basic-Car-Maintenance/assets/18750749/005af7ad-499a-46aa-a531-1ecf0dc78aaf

n recording:

### DIFF
--- a/Basic-Car-Maintenance/Shared/Settings/Views/SettingsView.swift
+++ b/Basic-Car-Maintenance/Shared/Settings/Views/SettingsView.swift
@@ -250,7 +250,7 @@ struct SettingsView: View {
                     showDeleteVehicleAlert = false
                 }
             } message: {
-                Text("The last vehicle can't be deleted", comment: "Alert message preventing users from deleting their last vehicle")
+                Text("The last vehicle can't be deleted. Please add a new vehicle before removing this one.", comment: "Alert message preventing users from deleting their last vehicle")
             }
             .navigationTitle(Text("Settings", comment: "Label to display settings."))
             .task {
--- a/Basic-Car-Maintenance/Shared/Settings/Views/SettingsView.swift
+++ b/Basic-Car-Maintenance/Shared/Settings/Views/SettingsView.swift
@@ -19,6 +19,7 @@ struct SettingsView: View {
     @State private var viewModel: SettingsViewModel
     @State private var isShowingAddVehicle = false
     @State private var showDeleteVehicleError = false
+    @State private var showDeleteVehicleAlert = false
     @State private var showAddVehicleError = false
     @State private var errorDetails: Error?
     @State private var copiedAppVersion: Bool = false
@@ -117,7 +118,11 @@ struct SettingsView: View {
                             Button(role: .destructive) {
                                 Task {
                                     do {
-                                        try await viewModel.deleteVehicle(vehicle)
+                                        if viewModel.vehicles.count > 1 {
+                                            try await viewModel.deleteVehicle(vehicle)
+                                        } else {
+                                            showDeleteVehicleAlert = true
+                                        }
                                     } catch {
                                         errorDetails = error
                                         showDeleteVehicleError = true
@@ -239,6 +244,13 @@ struct SettingsView: View {
                     Text("Failed To Delete Vehicle. Unknown Error.",
                          comment: "Label to display error details.")
                 }
+            }
+            .alert("Can't Delete Last Vehicle", isPresented: $showDeleteVehicleAlert) {
+                Button("OK", role: .cancel) {
+                    showDeleteVehicleAlert = false
+                }
+            } message: {
+                Text("The last vehicle can't be deleted", comment: "Alert message preventing users from deleting their last vehicle")
             }
             .navigationTitle(Text("Settings", comment: "Label to display settings."))
             .task {
--- a/Basic-Car-Maintenance/Shared/Localizable.xcstrings
+++ b/Basic-Car-Maintenance/Shared/Localizable.xcstrings
@@ -3950,22 +3950,8 @@
         }
       }
     },
-    "The last vehicle can't be deleted" : {
-      "comment" : "Alert message preventing users from deleting their last vehicle",
-      "localizations" : {
-        "ru" : {
-          "stringUnit" : {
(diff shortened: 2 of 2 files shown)


windrunner21:
@mikaelacaron I have also added some localization options for Turkish and Russian languages for the Alert I have created. I know English, Azerbaijani, Russian and Turkish languages. Also have limited knowledge in Japanese and French. If any help is needed in localization part, feel free to ask as well!

windrunner21:
@mikaelacaron hey, just checking up on you 😁 anything you need?

mikaelacaron:
@windrunner21 Thanks! Sorry I haven't had a chance to look at this yet! I'll get to it this weekend

windrunner21:
Sure, looking forward to your comments! Good luck in the meantime!

Jakayus:
@windrunner21  @mikaelacaron This is a suggestion, but I am wondering if we can improve UX with the alert message text  as the user might question why we can't delete the last vehicle. I used ChatGPT to quickly brainstorm some message versions and these seemed like the best.

"The last vehicle can't be deleted. Please ensure at least one vehicle remains in your account."
"You must have at least one vehicle in your account. Please add a new vehicle before deleting this one."
"Oops! Looks like this is your only vehicle. To keep using our services, you need at least one vehicle in your account."

Those are just suggestions/ideas! I also don't want to add a bunch of last minute work to the PR, especially considering localization efforts needed.

windrunner21:
@Jakayus hey, I agree with you as I was under the same impression while writing this feature. However, the reason I didn't decide to mention it to @mikaelacaron is that I think this feature is temporary, and this restriction may be removed in the future. If not, then it's not a problem for me, let's decide on the improved prompt and I'll add it! 

Good catch!

Jakayus:
@windrunner21 Thank you for the background!

I would say let's go with `The last vehicle can't be deleted. Please add a new vehicle before removing this one.` This already includes some of what you have previously, so the localization is already halfway there. It's more action oriented than reason oriented, not sure what we prefer for UX.

I think too if we want to use a fancier/friendlier prompt, perhaps @mikaelacaron can weigh in on specific wording. The above is sort of the quickest fix.

Basic-Car-Maintenance/Shared/Settings/Views/SettingsView.swift
```diff
@@ -240,6 +245,13 @@ struct SettingsView: View {
                          comment: "Label to display error details.")
                 }
             }
+            .alert("Can't Delete Last Vehicle", isPresented: $showDeleteVehicleAlert) {
+                Button("OK", role: .cancel) {
+                    showDeleteVehicleAlert = false
+                }
+            } message: {
+                Text("The last vehicle can't be deleted", comment: "Alert message preventing users from deleting their last vehicle")
```
mikaelacaron:
```suggestion
                Text("The last vehicle can't be deleted. Please add a new vehicle before removing this one.", comment: "Alert message preventing users from deleting their last vehicle")
```

mikaelacaron (commented):
Great work!!

I think this wording is best, I'm going to manually change it and push it to your branch
> The last vehicle can't be deleted. Please add a new vehicle before removing this one.

mikaelacaron:
> However, the reason I didn't decide to mention it to @mikaelacaron is that I think this feature is temporary, and this restriction may be removed in the future. 

All good! I'm totally fine with you (and anyone) bringing up any suggestions for how I've written any feature!

This issue, the reason for it right now is because if you try to add a maintenance event or odometer reading, without having a vehicle, the row only says "Select a vehicle" without anything else. And because the vehicle doesn't exist, the user would need to figure out that they have to go back to settings to add the vehicle.

Adding an onboarding flow #40 prevents this from happening (for this MVP), but the user can still delete vehicles, so the overall issue still exists. #202 prevents the user from deleting all vehicles so this isn't a problem

Later on after the 1st version is shipped I'm thinking of a longer term solution to this issue

mikaelacaron:
> I have also added some localization options for Turkish and Russian languages for the Alert I have created.

Oh also @windrunner21 when adding new features, don't worry about not being able to translate everything into all the different languages this app supports (if you do know a language the app supports, like Russian) you can go ahead and add the translation for the new string


If you would like to add a new language that isn't supported yet, you can comment on #7 what language you'd like to add, and then open a PR with only those changes

windrunner21:
> This issue, the reason for it right now is because if you try to add a maintenance event or odometer reading, without having a vehicle, the row only says "Select a vehicle" without anything else. And because the vehicle doesn't exist, the user would need to figure out that they have to go back to settings to add the vehicle.

@mikaelacaron I think we can display "Add vehicle" button instead of "Select a vehicle" row in this user story, in order to make it clear for the end user to add their vehicle first. If you think that this solutions makes sense, I can work on it if you'd like.

mikaelacaron:
> @mikaelacaron I think we can display "Add vehicle" button instead of "Select a vehicle" row in this user story, in order to make it clear for the end user to add their vehicle first. If you think that this solutions makes sense, I can work on it if you'd like.

This doesn't make sense right now because the user will try to tap on it, and still nothing happens

If you want to see this, in your simulator do "erase all settings" and then run the app, and you'll see

PR #280, opened 2023-12-29
Add Polish localization

maartinj:
# What it Does
* https://github.com/mikaelacaron/Basic-Car-Maintenance/issues/7
* App can use Polish translation

# How I Tested
* Run the application in simulator with language set to Polish
* Checked screens that show proper Polish translation

# Notes
* May need further updates as more features are developed where new strings will be available

# Screencast
https://github.com/mikaelacaron/Basic-Car-Maintenance/assets/53284922/ae5ae917-e161-4345-8c0b-626ccd67d0e5


mikaelacaron (approved):
Great work!! Thanks for adding another language!
yeah you're right as the app grows some strings will be out dated, but that will naturally happen with all of the languages. (until I look into different ways to translate them).

But until then, you're more than welcome to come back anytime and update the localization

maartinj:
Hi @mikaelacaron, thank you for your feedback! Sure, I will be happy to update the language when will be needed.
Happy New Year! 🎉🍾🎉

PR #300, opened 2024-08-03
fix: infinite loading on Dashboard

### LINKED ISSUE #288: BUG - DashboardView Loads Infinitly
## Describe the bug
* The `DashboardView` is always in the `isLoading` state because there is no error handling if the loading fails
* In `DashboardViewModel` function `getMaintenanceEvents` the line `let querySnapshot = try? await docRef.getDocuments()` doesn't handle any errors and it should

## To Reproduce
1. Run the app
2. The `DashboardView` loads with a `ProgressView`

## Expected behavior
* After there is an error show the empty screen, and make `isLoading` false, so that it isn't infinitely showing a `ProgeressView`

## Screenshots
* If there is an error set `isLoading` to `false` and show the empty view (which will be implemented in #286 

 * OS: iOS 17.5
 * App Version v1.0(1)

## Additional Context
* Don't make an empty view because that's in a different issue, change `isLoading to `false`

UVCoder1002:
# What it Does
* Closes #288 
* Fetching snapshot from firebase can throw error so putting the code inside do catch and handling the case of error by setting loading state to false

# How I Tested
* Run the app
* The DashboardView loads with a ProgressView for a finite time.

# Screenshot
Now screen loads finitely in case of any error too.
https://github.com/user-attachments/assets/fdc035ac-11bd-438a-b8e3-3297ae50c39e

### DIFF
--- a/Basic-Car-Maintenance/Shared/Dashboard/ViewModels/DashboardViewModel.swift
+++ b/Basic-Car-Maintenance/Shared/Dashboard/ViewModels/DashboardViewModel.swift
@@ -75,23 +75,23 @@ class DashboardViewModel {
 
         if let userUID = userUID {
             let db = Firestore.firestore()
-            
-            let docRef = db.collectionGroup(FirestoreCollection.maintenanceEvents)
-                .whereField(FirestoreField.userID, isEqualTo: userUID)
-            
-            let querySnapshot = try? await docRef.getDocuments()
-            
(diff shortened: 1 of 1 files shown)


mikaelacaron (approved):
Great work!! Thanks for this! I'll be getting the repo ready for Hacktoberfest 2024 soon!

(I'll merge this at my computer, currently on my phone).

PR #302, opened 2024-08-30
Implement AddOdometerReading AppIntent

### LINKED ISSUE #222: FEATURE - Make an App Intent for Adding an Odometer Reading with Shortcuts

OmarHegazy93:
# What it Does
* Closes #222
Implement shortcut for adding new odometer reading

### What is implemented?
1. `Vehicle` now conforms to `AppEntity` protocol, hence, it should have an explicit `id`, which represents `documentId` under the hood
2. `VehicleQuery` is simply fetching vehicles from the server
3. `DistanceUnit` enum is used to make it more readable for selecting Distance unit
4. In `AddOdometerReadingIntent`, validation is applied to `Distance`, and if the vehicles are empty, then the user will not be able to finish the shortcut

# How I Tested
###  Happy Scenario
1. Install the app
2. Make sure at least one vehicle is added in the app by going to `Settings` tab > `Add Vehicle`
3. Switch to Shortcuts app
4. Tab on the `+` on the top right corner
5. Tab on `⨁ Add Action`
6. Select `Apps` tab
7. Tab on `Basic Car`
8. Tab on `Add Odometer Reading`
9. Add data for all fields and tab the play button on the bottom right corner
10. See the result

#### Senario1
1. Install the app
2. Make sure at least one vehicle is added in the app by going to `Settings` tab > `Add Vehicle`
3. Switch to Shortcuts app
4. Tab on the `+` on the top right corner
5. Tab on `⨁ Add Action`
6. Select `Apps` tab
7. Tab on `Basic Car`
8. Tab on `Add Odometer Reading`
9. Try to set distance value to zero or negative value
10. See the result (user won't be able to proceed if the distance value is less than 1)

#### Senario2
1. Install the app
2. Make sure **no** vehicles are added
3. Switch to Shortcuts app
4. Tab on the `+` on the top right corner
5. Tab on `⨁ Add Action`
6. Select `Apps` tab
7. Tab on `Basic Car`
8. Tab on `Add Odometer Reading`
9. Try to set distance value to zero or negative value
10. See the result (user won't be able to proceed because there are no vehicles)


Basic-Car-Maintenance/Shared/AppIntents/AddOdometerReadingAppIntent.swift
```diff
@@ -0,0 +1,90 @@
...
+            .mile: "Miles",
+            .kilometer: "Kilometers"
+        ]
+    }
+    
+    case mile
+    case kilometer
+}
+
+struct AddOdometerReadingIntent: AppIntent {
```
mikaelacaron: How does this work when it's not called anywhere? Is there a WWDC video I could reference to look at this?

Basic-Car-Maintenance/Shared/AppIntents/AddOdometerReadingAppIntent.swift
```diff
@@ -0,0 +1,90 @@
...
+        let reading = OdometerReading(
+            date: date,
+            distance: distance,
+            isMetric: distanceType == .kilometer,
+            vehicleID: vehicle.id
+        )
+        let authViewModel = await AuthenticationViewModel()
+        let odometerVM = OdometerViewModel(userUID: authViewModel.user?.uid)
+        try odometerVM.addReading(reading)
+        return .result(dialog: "Added reading successfully")
```
mikaelacaron: I see why you need this, but I don't like having this here in this struct because it makes it impossible to test this, but also because now `AddOdometerReadingIntent` depends on `AuthenticationViewModel` and `OdometerViewModel`

I'm not sure what the alternative is, which kinda goes back to my question about when is this struct called?

Basic-Car-Maintenance/Shared/AppIntents/AddOdometerReadingAppIntent.swift
```diff
@@ -0,0 +1,90 @@
...
+    
+    var localizedStringResource: LocalizedStringResource {
+        switch self {
+        case .invalidDistance:
+            "You can not select distance number less than 1 km or mi"
+        case .emptyVehicles:
+            "No vehicles available, please add a vehicle using the app and try again"
+        }
+    }
+}
```
mikaelacaron: can you document all of this, using `///` comments so it'll show up in DocC

Basic-Car-Maintenance/Shared/AppIntents/AddOdometerReadingAppIntent.swift
```diff
@@ -0,0 +1,90 @@
...
+    static var caseDisplayRepresentations: [DistanceUnit: DisplayRepresentation] {
+        [
+            .mile: "Miles",
+            .kilometer: "Kilometers"
+        ]
+    }
+    
+    case mile
+    case kilometer
+}
```
mikaelacaron: also comment this with `///` for DocC

Basic-Car-Maintenance/Shared/Localizable.xcstrings
```diff
@@ -368,6 +368,9 @@
           }
         }
       }
+    },
```
mikaelacaron, line 348: all of the new localized strings should also have comments

Basic-Car-Maintenance/Shared/Models/Vehicle.swift
```diff
@@ -7,9 +7,13 @@
...
 import FirebaseFirestoreSwift
 import Foundation
+import AppIntents
 
-struct Vehicle: Codable, Identifiable, Hashable {
-    @DocumentID var id: String?
+struct Vehicle: Codable, Identifiable, Hashable, AppEntity {
+    var id: String {
+        documentID ?? ""
+    }
```
mikaelacaron: put this conformance in an extension of `Vehicle`

Basic-Car-Maintenance/Shared/Models/Vehicle.swift
```diff
@@ -18,17 +22,23 @@ struct Vehicle: Codable, Identifiable, Hashable {
...
-         name: String,
-         make: String,
-         model: String,
-         year: String? = nil,
-         color: String? = nil,
-         vin: String? = nil,
-         licensePlateNumber: String? = nil) {
-        self.id = id
+    init(
+        id: String = "",
```
mikaelacaron: This should not be set like this, it doesn't need to be here

Basic-Car-Maintenance/Shared/Models/Vehicle.swift
```diff
@@ -18,17 +22,23 @@ struct Vehicle: Codable, Identifiable, Hashable {
...
+        userID: String? = nil,
+        name: String,
+        make: String,
+        model: String,
+        year: String? = nil,
+        color: String? = nil,
+        vin: String? = nil,
+        licensePlateNumber: String? = nil
+    ) {
+        self.documentID = id
```
mikaelacaron, line 41: The documentID shouldn't be set from the id, because it would always be empty

this gets set properly when fetching from Firebase

Basic-Car-Maintenance/Shared/Settings/ViewModels/SettingsViewModel.swift
```diff
@@ -82,14 +82,13 @@ final class SettingsViewModel {
...
         if let userUID = authenticationViewModel.user?.uid {
-            guard let vehicleID = vehicle.id else { return }
             var vehicleToUpdate = vehicle
             vehicleToUpdate.userID = userUID
             
             do {
                 try Firestore.firestore()
                     .collection(FirestoreCollection.vehicles)
-                    .document(vehicleID)
+                    .document(vehicle.id)
```
mikaelacaron, line 92: What is the reason for not checking `vehicle.id` isn't nil?

mikaelacaron (changes requested):
Thank you for submitting this!
I have several comments about the implementation, and a few questions

Basic-Car-Maintenance/Shared/Models/Vehicle.swift
OmarHegazy93: So there are a couple of things going on here:
1. Since `Vehicle` conforms now to `AppEntity` protocol, it must be `identifiable`, meaning it must have a __non-optional__ `id`
2. The property `documentID` was previously named `id` and since it's defined using `@DocumentID` property wrapper, it must be an optional value, so by keeping it as `id` will give a compilation error indicating that Type `Vehicle` does not conform to protocol `AppEntity`, which must be non-optional as motioned in the aforementioned point
3. The `documentID` will correspond to the key "_id" when sending/retrieving vehicle from firebase as the `CodingKeys` enum indicates
4. The reason why `documentID` is set to private property is not to confuse the caller for having `documentID` and `id` so I wanted the caller to deal with only one value that refers to the id of the vehicle and under the hood, we pass this id to the `documentID` and when calling `vehicle.id`, this implicitly get the `documentID` 

Hopefully that made it clear

Basic-Car-Maintenance/Shared/Settings/ViewModels/SettingsViewModel.swift
OmarHegazy93: because `id` is now a non-optional value, and even when adding a default value to the vehicle id if it doesn't have an id, that should never happen in this case because when updating an existing vehicle, it must have an id

Basic-Car-Maintenance/Shared/AppIntents/AddOdometerReadingAppIntent.swift
OmarHegazy93: It's not about calling that inside the app, it's about what we expose to the system so it can use to make the user interacts with it without the need to open the app.

but basically what happens here is:
1. The system now knows about `AddOdometerReadingIntent` as it conforms to `AppIntent` protocol, so when for instance, when searching for the actions related to this app in the shortcuts app, you will automatically find this one with the title "Add Odometer Reading"
2. we have a set of `Parameters` for the inputs we want to enter so that we can add the reading
3. once all parameters are entered correctly, then the method `perform()` is called to perform that app intent

you can refer to [this](https://developer.apple.com/videos/play/wwdc2022/10032) WWDC session for more info

Basic-Car-Maintenance/Shared/AppIntents/AddOdometerReadingAppIntent.swift
OmarHegazy93: let me know if the previous answer addresses your question, but here the only way we can access the user id to retrieve the Vehicles is through `AuthenticationViewModel`, however, one alternative solution would be persisting the vehicles locally on disk so that we can access vehicles from inside `OdometerViewModel`, hence, no need for the `AuthenticationViewModel` dependency anymore nor calling an api to fetch the vehicles from the server. 
Or maybe storing the UserID somewhere and make it accessible inside `OdometerViewModel, but this one wouldn't make sense to me

mikaelacaron:
@OmarHegazy93 please click "resolve" on all the issues that you've addressed

OmarHegazy93:
Sorry @mikaelacaron , I re-requested the review and forgot to submit the comments 😄, you should be able to see them now

Basic-Car-Maintenance/Shared/AppIntents/AddOdometerReadingAppIntent.swift
```diff
@@ -0,0 +1,137 @@
...
+        try await fetchVehicles()
+    }
+    
+    func suggestedEntities() async throws -> [Vehicle] {
+        try await fetchVehicles()
+    }
+    
+    private func fetchVehicles() async throws -> [Vehicle] {
+        let authViewModel = await AuthenticationViewModel()
+        let odometerVM = OdometerViewModel(userUID: authViewModel.user?.uid)
```
mikaelacaron: So from what I've seen in examples we'd use `@Dependency` for these dependencies in the `EntityQuery` and `AppIntent`

Like:
`@Dependency private var authViewModel: AuthenticationViewModel` as a struct level variable, then we use it in perform and `fetchVehicles`, rather than initializing it

Can you add that here and in the other places too

mikaelacaron (changes requested):
Just another comment, thanks for all the work on this! Can you also update this branch from change on `dev`

Basic-Car-Maintenance/Shared/AppIntents/AddOdometerReadingAppIntent.swift
mikaelacaron: @OmarHegazy93

Basic-Car-Maintenance/Shared/Settings/ViewModels/AuthenticationViewModel.swift
```diff
@@ -24,7 +24,7 @@ enum AuthenticationFlow {
 }
 
 @Observable
-final class AuthenticationViewModel {
+final class AuthenticationViewModel: @unchecked Sendable {
```
mikaelacaron, line 27: this is the warning in the `AddOdometerReadingIntent` file, if we don't have this

Stored property '_authViewModel' of 'Sendable'-conforming struct 'AddOdometerReadingIntent' has non-sendable type 'AppDependency<AuthenticationViewModel>'; this is an error in the Swift 6 language mode

Basic-Car-Maintenance/Shared/AppIntents/AddOdometerReadingIntent/AddOdometerReadingIntent.swift
```diff
@@ -0,0 +1,97 @@
...
+    )
+    var distance: Int
+        
+    static var title = LocalizedStringResource(
+        "Add Odometer Reading",
+        comment: "Title for the app intent when adding an odometer reading"
+    )
+    
+    private func fetchVehicles() async throws -> [Vehicle] {
+        let odometerVM = OdometerViewModel(userUID: authViewModel.user?.uid)
```
mikaelacaron, line 55: let's add `OdometerViewModel` as a dependency too

mikaelacaron (commented):
we reviewed this live during my livestream!

https://youtube.com/live/xhlq3p9gxDM?feature=share

we didn't finish it, but there's a few comments of things we are going to change!

PR #337, opened 2024-10-10
Update Firestore Rules, Update Packages

mikaelacaron:
# What it Does
* Update Firestore rules in the docs and the emulator to only be able to edit your own data and not others
* Update packages (Firebase version 12+)
* Remove deprecated libraries
   * FirebaseRemoteConfigSwift
   * FirebaseAnalyticsSwift
   * FirebaseFirestoreSwift
* 

# How I Tested
* I haven't, but in theory, these things should still work: add a vehicle, maintenance event, and odometer reading


mikaelacaron:
Firebase is updated, and rules are written in docs, but on this branch the read write is failing, even when the rules are open with `allow read, write: if true;`, which could be a bug with Firebase? I'll merge this after chatting with Firebase support

PR #338, opened 2024-10-10
Added app icons as a preview in ChooseAppIcon view

### LINKED ISSUE #330: BUG - Icons preview in ChooseAppIconView not visible
## Describe the bug
* As mentioned in the topic of the bug, all alternative icons are not visible in `ChooseAppIconView` for some reason

## To Reproduce
1. Run the app
2. Tap `SettingsView`
3. Tap `Change App Icon`
4. See icons preview not visible

## Expected behavior
* Show all alternative icons in `ChooseAppIconView`

## Device Info:
* OS: iOS 17.5
* App Version v1.0(1)

mitaligondaliya:
# What it Does
* Fixes #330 
* Added Image Sets for App Icon Previews:
* Created separate Image Sets for the primary and alternate app icons to display within the app.
* Updated the ChooseAppIconView to now correctly show previews of the app icons using the newly added image sets.

# How I Tested
For example:
* Run the application
* Go to the Settings tab
* Tap the Choose App Icon button

# Notes
*  This approach replaces the deprecated method of fetching app icons directly from the App Icon set using UIImage(named:).
* This ensures reliable in-app display of both the primary and alternate icons, which was not working with the CFBundleIcons method.

# Screenshot
Simulator Screenshot - iPhone 16 Pro - 2024-10-10 at 18 25 50


mikaelacaron (changes requested):
@mitaligondaliya good work! the app icons show on the `ChooseAppIconView`, but it looks like this solution breaks the tinted and dark colored icons

is it possible to only add the IconPreviews folder, and not change the app icon types? (so it's only light, dark and tinted, as options rather than all the scale options)

and the icon previews, can we set the scale to "Single Scale"?

Screenshot 2024-10-14 at 1 26 38 PM

mitaligondaliya:
> is it possible to only add the IconPreviews folder, and not change the app icon types? (so it's only light, dark and tinted, as options rather than all the scale options)

What do you mean by this? Are you referring to supporting dark and tinted icons? If so, we can add that. However, as I can see, we haven't supported that in the current app icon set.

mitaligondaliya:
> and the icon previews, can we set the scale to "Single Scale"?

Yes we can set that.

mikaelacaron:
> > is it possible to only add the IconPreviews folder, and not change the app icon types? (so it's only light, dark and tinted, as options rather than all the scale options)
> 
> What do you mean by this? Are you referring to supporting dark and tinted icons? If so, we can add that. However, as I can see, we haven't supported that in the current app icon set.


yeah right now it does! you may need to update your fork to see it
Screenshot 2024-10-14 at 9 17 53 PM

mikaelacaron:
@mitaligondaliya can you clarify again, I'm kinda confused, why do we need to duplicate all of the app icon assets?? When the "Icon Previews" are all also 1024x1024? it looks like according to the code change, nothing is needed except using the `rawValue` to call the icon name to set it in SwiftUI, rather than an `Image`, what does duplicating all the icons do??

Screenshot 2024-10-16 at 3 56 03 PM

mitaligondaliya:
> @mitaligondaliya can you clarify again, I'm kinda confused, why do we need to duplicate all of the app icon assets?? When the "Icon Previews" are all also 1024x1024? it looks like according to the code change, nothing is needed except using the `rawValue` to call the icon name to set it in SwiftUI, rather than an `Image`, what does duplicating all the icons do??
> 
> Screenshot 2024-10-16 at 3 56 03 PM

@mikaelacaron From iOS 18 onward, alternate app icons are no longer directly accessible by SwiftUI views. If you want to display them within the app, they must be added separately as regular image assets in the asset catalog.

mikaelacaron:
How does the code know you're referencing the "Icon Previews" vs the app icon files? Cause it looks like regular app icons were added

mitaligondaliya:
> How does the code know you're referencing the "Icon Previews" vs the app icon files? Cause it looks like regular app icons were added

I've added the same icons because I don't have preview files available, but I can add them in the required size for icon previews and will change the name and code accordingly.

mikaelacaron:
> I've added the same icons because I don't have preview files available, but I can add them in the required size for icon previews.but I can add them in the required size for icon previews and will change the name and code accordingly.

what do you mean "preview files?" like the 64x64 size? Sorry I don't understand what needs to change vs what's in the PR right now?
I can add you to the Figma to download / make the sizes you need, DM me on any social media, and I can add you

mitaligondaliya:
> > I've added the same icons because I don't have preview files available, but I can add them in the required size for icon previews.but I can add them in the required size for icon previews and will change the name and code accordingly.
> 
> what do you mean "preview files?" like the 64x64 size? Sorry I don't understand what needs to change vs what's in the PR right now? I can add you to the Figma to download / make the sizes you need, DM me on any social media, and I can add you

Yes, 64x64 size. We need to fix the app icon display issue, as it's currently not showing. The changes I've added should resolve this, and we also need the proper size preview icons, for which I'll ping you.

mitaligondaliya:
@mikaelacaron I'm done with this fix. Kindly review it and let me know if any changes are needed. Thank you!

mikaelacaron (approved):
thanks so much for all the changes and explanations! great work!!

PR #341, opened 2024-10-15
Vehicle Display in settings

### LINKED ISSUE #332: IMPROVE -Vehicle Display in settings
Reformat how the vehicle details are displayed in the settings screen.

N/A
VehicleDisplay

maartinj:
# What it Does
* Closes #332 
* Changed how the vehicle details are displayed in the settings screen, to match proposition made by @rramirez-dev

# How I Tested
* Run the application in Simulator (iPhone 15 Pro, iOS 17.5), set language to 'English'
* Click tab 'Settings'
* Click row 'Add Vehicle' in 'VEHICLES' section
* Add information in NAME, MAKE, MODEL, YEAR, COLOR, VIN, LICENSE PLATE NUMBER section
* Click 'Add' button
* See the screen in 'Screenshot' section, where are displayed vehicle details in changed format

# Notes
* Changes in text also works when User will want to change vehicle details when vehicle details already exist

# Screenshot

### DIFF
--- a/Basic-Car-Maintenance/Shared/Settings/Views/SettingsView.swift
+++ b/Basic-Car-Maintenance/Shared/Settings/Views/SettingsView.swift
@@ -91,19 +91,21 @@ struct SettingsView: View {
                 Section {
                     ForEach(viewModel.vehicles) { vehicle in
                         VStack(alignment: .leading, spacing: 2) {
-                            HStack {
-                                if let year = vehicle.year, !year.isEmpty {
-                                    Text(year)
-                                }
-                                
-                                Text(vehicle.make)
-                                
-                                Text(vehicle.model)
-                            }
-                            .fontWeight(.bold)
-                            .font(.headline)
+                            Text("Display Name: \(vehicle.name)")
+                                .fontWeight(.bold)
+                                .font(.headline)
                             
                             Group {
+                                HStack {
+                                    if let year = vehicle.year, !year.isEmpty {
+                                        Text(year)
+                                    }
+                                    
+                                    Text(vehicle.make)
+                                    
+                                    Text(vehicle.model)
+                                }
+                                
                                 if let licensePlateNumber =
                                     vehicle.licensePlateNumber,
                                    !licensePlateNumber.isEmpty {
@@ -116,9 +118,7 @@ struct SettingsView: View {
                                 
                                 if let color = vehicle.color, !color.isEmpty {
                                     Text("Color: \(color)")
-                                }
-                                
-                                Text("Display Name: \(vehicle.name)")
+                                } 
                             }
                             .font(.callout)
                             .foregroundStyle(.secondary)
--- a/Basic-Car-Maintenance/Shared/Settings/Views/SettingsView.swift
+++ b/Basic-Car-Maintenance/Shared/Settings/Views/SettingsView.swift
@@ -29,6 +29,15 @@ struct SettingsView: View {
     @State private var isShowingEditVehicleView = false
     
     private let appVersion = "Version \(Bundle.main.versionNumber) (\(Bundle.main.buildNumber))"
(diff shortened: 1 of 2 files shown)


Basic-Car-Maintenance/Shared/Settings/Views/SettingsView.swift
```diff
@@ -90,29 +99,26 @@ struct SettingsView: View {
...
                             
-                            if let licensePlateNumber = vehicle.licensePlateNumber,
-                               !licensePlateNumber.isEmpty {
-                                Text(licensePlateNumber)
+                            ForEach(vehicleDetails(for: vehicle), id: \.label) { detail in
+                                if let value = detail.value, !value.isEmpty {
+                                    Text("\(detail.label): \(value)")
+                                        .font(.callout)
+                                        .foregroundStyle(.secondary)
+                                }
```
mikaelacaron: can we keep this closer to the original code, it is more repetitive, but it's clearer, as opposed to using the `vehicleDetails` loop or the year, make, model loop

Basic-Car-Maintenance/Shared/Settings/Views/SettingsView.swift
```diff
@@ -29,6 +29,15 @@ struct SettingsView: View {
     @State private var isShowingEditVehicleView = false
     
     private let appVersion = "Version \(Bundle.main.versionNumber) (\(Bundle.main.buildNumber))"
+
+    private func vehicleDetails(for vehicle: Vehicle) -> [(label: String, value: String?)] {
```
mikaelacaron: also note I would like to remove this, but if you were to do this, the label should be a `LocalizedStringResource` because that makes it accessible, so that this all could be translated

mikaelacaron (changes requested):
Great work!! I have several comments that need resolved please. If you have any questions, you can comment directly on that thread, otherwise click the resolve button once you've addressed my comment. Once you fix everything, please click the re-review button
Re-review button

and I was looking at the design, can we actually make the vehicle name the bolded words, as opposed to the make / model

maartinj:
> Great work!! I have several comments that need resolved please. If you have any questions, you can comment directly on that thread, otherwise click the resolve button once you've addressed my comment. Once you fix everything, please click the re-review button Re-review button
> 
> and I was looking at the design, can we actually make the vehicle name the bolded words, as opposed to the make / model

Something like this?



\
I corrected PR with prefix v2 and above change about changing bolded words with new commit. Please check and let me know 👍

Basic-Car-Maintenance/Shared/Settings/Views/SettingsView.swift
maartinj: Changed as proposed, please let me know 👍

Basic-Car-Maintenance/Shared/Settings/Views/SettingsView.swift
maartinj: Changed as proposed - deleted func vehicleDetails. Please let me know 👍

Basic-Car-Maintenance/Shared/Settings/Views/SettingsView.swift
```diff
@@ -90,30 +90,38 @@ struct SettingsView: View {
                 
                 Section {
                     ForEach(viewModel.vehicles) { vehicle in
-                        VStack(alignment: .leading) {
-                            Text(vehicle.name)
+                        VStack(alignment: .leading, spacing: 2) {
+                            Text("Display Name: \(vehicle.name)")
```
mikaelacaron: I don't think we need the word "Display Name" just using the name is fine

mikaelacaron (approved):
Great work!! Thanks for contributing!

PR #344, opened 2024-10-15
Add Persian localization

nssina:
# What it Does
* Added Persian localization 

# How I Tested
* Run the application
* Go to device Settings
* Find the app settings
* Change language to Persian

# Notes
We're missing translation key for search bar placeholder. Currently it's "Search" which will translate to "جستجو" in Persian.

# Screenshot
IMG_0001 IMG_0001


mikaelacaron (approved):
Great work thanks!!
haha I found the reason Search isn't localized is because of `.searchable(text: $viewModel.searchText)` without the `prompt` which is a default parameter, it doesn't get localized. I'll add this

PR #345, opened 2024-10-19
Implement export maintenance events

### LINKED ISSUE #342: FEATURE - PDF Export of Maintenance Events Only
* Create a PDF export of the maintenance events ordered by oldest to newest of a single vehicle
* Add a toolbar button to start the export with `square.and.arrow.up` on the leading side of the `DashboardView`
   * Tapping this shows `ExportOptionsView` as a sheet
* Create a view called `ExportOptionsView` (in the `Dashboard` > `Views` directory) that's used to eventually select the options for how the PDF can be exported (making it more customizable, but for the start we'll only do specific things, listed below)
   * Select the vehicle to export using a picker
   * Export button

Full flow:
* Tap export button
* See `ExportOptionsView` as a sheet, it'll show a picker of the vehicles
* Tap an Export button (make this the trailing toolbar button of this view
* Export as described below as a PDF, that can be saved to a person's file's app or AirDropped to someone

* N/A, new features will be added as this feature evolves
* Note this feature will "compete with" #343 for whichever feature gets completed first, because they have very similar functionality, so whoever completes it first, will have merge conflicts with the other

## Additional context
Example PDF report includes:
* Basic Car Maintenance Title
* Vehicle Name at the top
* Start and end date of all the `maintenance_events` that get exported
* Vehicle name (yes for this specific issue it'll be repeating for now)
* Notes of the maintenance events

maintenance event export

OmarHegazy93:
# What it Does
* Closes #342
1. Make the user able to export the maintenance events for a specific vehicle
2. Sort the records based on the date ascendingy

# How I Tested
1. Run the App
2. Make sure that there is at least one vehicle added
3. Add a maintenance event from the dashboard screen
4. The share button will appear at the top left corner, tap on it
5. Select the vehicle you want to export the report for
6. See the result


mikaelacaron:
@OmarHegazy93 can you also please add a screenshot for this PR, and add back the heading, be sure not to delete the PR template headings

can you also write some more about how this code works, and why it's written the way that it's written, there's several new protocols, what are they for? what do they do? etc?

Basic-Car-Maintenance/Shared/Dashboard/Views/CarMaintenancePDFGenerator.swift
```diff
@@ -0,0 +1,230 @@
...
+//  https://github.com/mikaelacaron/Basic-Car-Maintenance
+//  See LICENSE for license information.
+//
+
+import UIKit
+import SwiftUI
+
+protocol PDFGeneratable {
+    func generatePDF() -> URL?
+}
```
mikaelacaron: what does this protocol do, when it's only implemented in one place?

Basic-Car-Maintenance/Shared/Dashboard/Views/CarMaintenancePDFGenerator.swift
```diff
@@ -0,0 +1,230 @@
...
+    // Define the PDF page size (A4 size in points)
+    private let pageWidth: CGFloat = 595.2
+    private let pageHeight: CGFloat = 841.8
+    private var pageSize: CGSize
+    
+    // Define page margins
+    private let topMargin: CGFloat = 50
+    private let bottomMargin: CGFloat = 50
+    private let leftMargin: CGFloat = 20
+    private let rightMargin: CGFloat = 20
```
mikaelacaron: are these page margins specific to this PDF size? like will it be different if we used different page sizes?

if they're different, we should probably throw this into the `PageDimension` enum I mentioned

Basic-Car-Maintenance/Shared/Dashboard/Views/CarMaintenancePDFGenerator.swift
```diff
@@ -0,0 +1,230 @@
...
+}
+
+final class CarMaintenancePDFGenerator: PDFGeneratable {
+    private let title = "Basic Car Maintenance" 
+    private let vehicleName: String
+    private let events: [MaintenanceEvent]
+    // Define the PDF page size (A4 size in points)
+    private let pageWidth: CGFloat = 595.2
+    private let pageHeight: CGFloat = 841.8
+    private var pageSize: CGSize
```
mikaelacaron: let's make this into an enum in utilities, something like:

```swift
enum PageDimension {
   case a4

   var pageWidth: CGFloat {
      switch self {
         case .a4:
            return 595.2
      }
   }

   // ... continue for `pageHeight`
}
```

Basic-Car-Maintenance/Shared/Dashboard/Views/DashboardView.swift
```diff
@@ -149,13 +152,51 @@ struct DashboardView: View {
                     }
                 }
             }
+            .toolbar {
+                if !viewModel.events.isEmpty {
+                    ToolbarItem(placement: .topBarLeading) {
+                        Button {
+                            isShowingShareView = true
```
mikaelacaron: this should probably be `isShowingExportOptionsView`

Basic-Car-Maintenance/Shared/Dashboard/Views/DashboardView.swift
```diff
@@ -149,13 +152,51 @@ struct DashboardView: View {
...
+                        } label: {
+                            Image(systemName: SFSymbol.share)
+                        }
+                        .accessibilityShowsLargeContentViewer {
+                            Label {
+                                Text("ExportEvent", comment: "Label for exporting maintenance events")
```
mikaelacaron: should there be a space in `ExportEvent`?

Basic-Car-Maintenance/Shared/Dashboard/Views/ExportOptionsView.swift
```diff
@@ -0,0 +1,70 @@
...
+    func fetchEvents(for vehicle: Vehicle) -> [MaintenanceEvent]
+}
+
+struct ExportOptionsView: View {
+    @Environment(\.dismiss) var dismiss
+    @State private var selectedVehicle: Vehicle?
+    @State private var isShowingShareSheet = false
+    private let eventsFetcher: MaintenanceEventsFetcher
+    private let onExport: (URL) -> Void
+    private let vehicles: [Vehicle]
```
mikaelacaron: can you add line breaks between environment, state, and then regular variables?

Basic-Car-Maintenance/Shared/Dashboard/Views/ExportOptionsView.swift
```diff
@@ -0,0 +1,70 @@
...
+        self.onExport = onExport
+        self._selectedVehicle = State(initialValue: vehicles.first)
+    }
+    
+    var body: some View {
+        NavigationView {
+            VStack(alignment: .leading, spacing: 16) {
+                Text("Select the vehicle you want to export the maintenance events for:")
+                    .font(.headline)
+                    .padding(.top, 20)
```
mikaelacaron, line 30: this could be the `title` of the `NavigationView` as opposed to just text in a `VStack`, because this won't work when the dynamic type size is really big

Basic-Car-Maintenance/Shared/Dashboard/Views/ExportOptionsView.swift
```diff
@@ -0,0 +1,70 @@
...
+                
+                Picker("Select a Vehicle", selection: $selectedVehicle) {
+                    ForEach(vehicles, id: \.id) { vehicle in
```
mikaelacaron: the `id: \.id` isn't needed because `Vehicle` is `Identifiable`

Basic-Car-Maintenance/Shared/Dashboard/Views/ExportOptionsView.swift
```diff
@@ -0,0 +1,70 @@
...
+                        Text(vehicle.name)
+                            .tag(vehicle as Vehicle?)
```
mikaelacaron: this can just be `.tag(vehicle)` without the typecasting, because `Vehicle` is `Hashable`

Basic-Car-Maintenance/Shared/Dashboard/Views/ExportOptionsView.swift
```diff
@@ -0,0 +1,70 @@
...
+                    }
+                }
+                .pickerStyle(InlinePickerStyle())
```
mikaelacaron:
```suggestion
                .pickerStyle(.inline)
```
This syntax can be changed to this

Basic-Car-Maintenance/Shared/Dashboard/Views/ExportOptionsView.swift
```diff
@@ -0,0 +1,70 @@
...
+//
+//  https://github.com/mikaelacaron/Basic-Car-Maintenance
+//  See LICENSE for license information.
+//
+
+import SwiftUI
+
+protocol MaintenanceEventsFetcher {
+    func fetchEvents(for vehicle: Vehicle) -> [MaintenanceEvent]
+}
```
mikaelacaron: can we actually remove this protocol, cause we're really passing the whole `DashboardViewModel`, which isn't necessary, to just use a single function

We could only have one function in the `DashboardViewModel`, that does the same thing, called `filterMaintenanceEventsForExport(for vehicle: Vehicle)`

haha it's a little wordy, but I think it works, you could also add `///` comments that describe what the function is for

Basic-Car-Maintenance/Shared/Dashboard/Views/PDFShareController.swift
```diff
@@ -0,0 +1,49 @@
...
+//  Basic-Car-Maintenance
+//
+//  https://github.com/mikaelacaron/Basic-Car-Maintenance
+//  See LICENSE for license information.
+//
+
+import UIKit
+import SwiftUI
+
+struct PDFShareController: UIViewControllerRepresentable {
```
mikaelacaron: I don't think we need this, cause we can use `ShareLink`, which calls the share sheet (the `UIActivityViewController`)

Basic-Car-Maintenance/Shared/Dashboard/Views/CarMaintenancePDFGenerator.swift
```diff
@@ -0,0 +1,230 @@
...
+    init(vehicleName: String, events: [MaintenanceEvent]) {
+        self.vehicleName = vehicleName
+        self.events = events
+        self.pageSize = CGSize(width: pageWidth, height: pageHeight)
+        self.columnWidth = (pageWidth - leftMargin - rightMargin) / 3 
+    }
+    
+    func generatePDF() -> URL? {
+        guard !events.isEmpty else { return nil }
+        let fileName = "\(vehicleName)MaintenanceReport.pdf"
```
mikaelacaron:
```suggestion
        let fileName = "\(vehicleName)-MaintenanceReport.pdf"
```

Basic-Car-Maintenance/Shared/Dashboard/Views/CarMaintenancePDFGenerator.swift
```diff
@@ -0,0 +1,230 @@
...
+        
+        // Get the path to the documents directory
```
mikaelacaron: please remove all these unnecessary comments, if they aren't self explanatory in the code

also because of these comments this looks like it was generated by ChatGPT, that's fine, as long as the code is readable, works, there's no SwiftLint warnings, and it's using the latest APIs (which is something ChatGPT is bad at because models aren't updated)

Basic-Car-Maintenance/Shared/Dashboard/Views/CarMaintenancePDFGenerator.swift
```diff
@@ -0,0 +1,230 @@
...
+        let pdfRenderer = UIGraphicsPDFRenderer(bounds: CGRect(origin: .zero, size: pageSize))
+        
+        // Render content to the PDF file
+        let pdfData = pdfRenderer.pdfData { context in
+            
+            // Initialize the y-position (start after the headers)
+            var yPosition: CGFloat = topMargin
+            
+            // Function to begin a new page
+            func beginNewPage(isFirstPage: Bool) {
```
mikaelacaron: this should be outside this, I don't like it being nested, even though it works

Basic-Car-Maintenance/Shared/Dashboard/Views/CarMaintenancePDFGenerator.swift
```diff
@@ -0,0 +1,230 @@
...
+                event.notes.draw(in: notesRect, withAttributes: tableRowAttributes)
+                
+                // Adjust the spacing for the next row
+                yPosition += 60
+            }
+        }
+        
+        // Save the PDF data to the file URL
+        do {
+            try pdfData.write(to: fileURL!)
```
mikaelacaron: don't force unwrap

Basic-Car-Maintenance/Shared/Dashboard/Views/CarMaintenancePDFGenerator.swift
```diff
@@ -0,0 +1,230 @@
...
+            return nil
+        }
+    }
+    
+    // Helper function to draw header
+    private func drawHeader(
+        context: UIGraphicsPDFRendererContext,
+        yPosition: inout CGFloat
+    ) {
+        let titleAttributes: [NSAttributedString.Key: Any] = [
```
mikaelacaron: a better comment would be `Draw the center header and header columns` because it's actually doing all of that, also don't call things "helper functions"

and just make this on one line, rather than 3 for the `drawHeader` and parameters

Basic-Car-Maintenance/Shared/Dashboard/Views/CarMaintenancePDFGenerator.swift
```diff
@@ -0,0 +1,230 @@
...
+        // Draw date range centered
+        dateRangeString.draw(at: CGPoint(x: dateRangeX, y: yPosition), withAttributes: subtitleAttributes)
+        yPosition += 50  // Add more space after the headers
+        
+        drawColumnsHeaders(yPosition: &yPosition)
+    }
+    
+    private func drawColumnsHeaders(
+        yPosition: inout CGFloat
+    ) {
```
mikaelacaron: this should also be on one line

Basic-Car-Maintenance/Shared/Dashboard/Views/CarMaintenancePDFGenerator.swift
```diff
@@ -0,0 +1,230 @@
...
+        ]
+        
+        dateColumnHeader
+            .draw(
+                at: CGPoint(
+                    x: leftMargin,
+                    y: yPosition
+                ),
+                withAttributes: subtitleAttributes
+            )
```
mikaelacaron:
```suggestion
        dateColumnHeader
            .draw(at: CGPoint(x: leftMargin, y: yPosition),
                  withAttributes: subtitleAttributes
            )
```

can you reformat it like this

Basic-Car-Maintenance/Shared/Dashboard/Views/CarMaintenancePDFGenerator.swift
```diff
@@ -0,0 +1,230 @@
...
+        
+        vehicleColumnHeader
+            .draw(
```
mikaelacaron: same with this one

mikaelacaron (changes requested):
good work @OmarHegazy93 ! I've left a bunch of comments that are mainly around code formatting
If you could update it, that would be great!
If you have any questions, you can comment directly on that thread, as always, otherwise click the resolve button once you've addressed my comment. Once you fix everything, please click the re-review button

Basic-Car-Maintenance/Shared/Dashboard/Views/CarMaintenancePDFGenerator.swift
OmarHegazy93: I added it so that if we used it in a viewModel, we can mock it, but since it's currently used directly inside `ExportOptionsView` then yes, no need to have it

Basic-Car-Maintenance/Shared/Dashboard/Views/CarMaintenancePDFGenerator.swift
OmarHegazy93: I'm not sure if we would use different margins for different pages, so I think we can keep them here for now

Basic-Car-Maintenance/Shared/Dashboard/Views/CarMaintenancePDFGenerator.swift
OmarHegazy93: yes that was done with the help of ChatGPT, and coming to the point of using not updated model, there was another approach to achieve this by converting the SwiftUI view to a UIView, then render it inside a PDF file, which I think would be easier to draw the info we need using SwiftUI, but for the sake of the time I decided to look at it later

Basic-Car-Maintenance/Shared/Dashboard/Views/CarMaintenancePDFGenerator.swift
OmarHegazy93: Gotcha, I know it may not be common in iOS, however, it save some time trying to see where the method is used, instead of having it private inside the parent then search if it may be used in different places within class or the struct context

Basic-Car-Maintenance/Shared/Dashboard/Views/ExportOptionsView.swift
OmarHegazy93: you've a valid point, I did it using another way, using a dictionary that has `Vehicle` as a key and the relevant events as a value, this way, I have all the needed information that gets injected only once during the view initialization. let me know if that works

Basic-Car-Maintenance/Shared/Dashboard/Views/PDFShareController.swift
OmarHegazy93: Nice tip! 👍

Basic-Car-Maintenance/Shared/Dashboard/Views/ExportOptionsView.swift
OmarHegazy93: adding the title to the NavigationView makes the title truncated as shown below, also there is no way to style the Text passed to the navigation title to make it smaller as I'm getting the warning

> Only unstyled text can be used with navigationTitle(_:)

image

Basic-Car-Maintenance/Shared/Dashboard/Views/ExportOptionsView.swift
```diff
@@ -0,0 +1,79 @@
...
+                    .thumbnail(
+                        of: CGSize(
+                            width: UIScreen.main.bounds.width,
+                            height: UIScreen.main.bounds.height / 2),
+                        for: .mediaBox
+                    ) {
+                    ShareLink(item: url) {
+                        VStack {
+                            Image(uiImage: thumbnail)
+                            Label("Share", image: SFSymbol.share)
```
mikaelacaron:
```suggestion
                            Label("Share", systemImage: SFSymbol.share)
```

Should be `systemImage`, rather than `image`

Basic-Car-Maintenance/Shared/Dashboard/Views/ExportOptionsView.swift
```diff
@@ -0,0 +1,79 @@
...
+    
+    init(dataSource: [Vehicle: [MaintenanceEvent]]) {
+        self.dataSource = dataSource
+        self._selectedVehicle = State(initialValue: dataSource.first?.key)
+    }
+    
+    var body: some View {
+        NavigationView {
+            VStack(alignment: .leading, spacing: 16) {
+                Text("Select the vehicle you want to export the maintenance events for:")
```
mikaelacaron, line 28: this could be a scroll view to be able to accommodate dynamic type sizes

Basic-Car-Maintenance/Shared/Dashboard/Views/ExportOptionsView.swift
```diff
@@ -0,0 +1,79 @@
...
+//  See LICENSE for license information.
+//
+
+import SwiftUI
+import PDFKit
+
+struct ExportOptionsView: View {
+    @Environment(\.dismiss) var dismiss
+    @State private var selectedVehicle: Vehicle?
+    @State private var isShowingShareSheet = false
```
mikaelacaron: renamed to `isShowingThumbnail` because it's not technically showing the share sheet yet, the `ShareLink` does that

mikaelacaron (approved):
great work!! thanks for joining my stream for this review!

https://youtube.com/live/I_pXpNRryjI?feature=share

This PR was reviewed live on my YouTube channel

PR #367, opened 2024-10-30
Dutch translation

lonnekemennen:
# What it Does
* Completes Dutch translation


mikaelacaron (approved):
great work!! thanks for adding a new language!

PR #371, opened 2024-11-21
Implement CSV export of maintenance events

### LINKED ISSUE #343: FEATURE - CSV Export of Maintenance Events Only
* Note that this feature and #342 will "compete" with who makes it first, because they have very similar functionality, and are going to end up having a merge conflict
* Create a CSV export of the maintenance events ordered by oldest to newest of a single vehicle
* Add a toolbar button to start the export with `square.and.arrow.up` on the leading side of the `DashboardView`
* Tapping this shows `ExportOptionsView` as a sheet
* Create a view called `ExportOptionsView` (in the `Dashboard` > `Views` directory) that's used to eventually select the options for how the CSV can be exported (making it more customizable, but for the start we'll only do specific things, listed below)
* Select the vehicle to export using a picker
* Export button (trailing toolbar button)

Full flow:
* Tap export button
* See `ExportOptionsView` as a sheet, it'll show a picker of the vehicles
* Tap an Export button (make this the trailing toolbar button of this view
* Export as described below as a CSV, that can be saved to a person's file's app or AirDropped to someone

More features will be added to this after this issue and #342 are completed, stay tuned!

## Additional context

CSV format:
```csv
date, vehicle name, notes
```

mitaligondaliya:
# What it Does
* Closes #343 
* Added a picker to allow users to choose between PDF and CSV export options.
* Implemented functionality to generate a CSV file containing the maintenance events of selected vehicles.
* Integrated ShareLink to enable sharing of the generated CSV file.

# How I Tested
* Run the application
* Go to the dashboard and add a new maintenance event. This will enable the share button.
* Tap the share button to proceed to the vehicle selection screen.
* Select a vehicle, and a picker will appear.
* Choose CSV as the export format.
* See the result

# Screenshot

https://github.com/user-attachments/assets/fed33098-55a7-4e22-8ad3-2c5011388af2

### DIFF
--- a/Basic-Car-Maintenance/Shared/Dashboard/Views/CSVEncoder.swift
+++ b/Basic-Car-Maintenance/Shared/Dashboard/Views/CSVEncoder.swift
@@ -1,0 +1,259 @@
+//
+//  CSVEncoder.swift
+//  Basic-Car-Maintenance
+//
+//  https://github.com/mikaelacaron/Basic-Car-Maintenance
+//  See LICENSE for license information.
+//
+
+import Foundation
+import SwiftUI
+
+struct CSVFile {
+    var events: [MaintenanceEvent]
+    
+    func csvData() -> String {
+        let table = CSVTable<MaintenanceEvent>(
+            columns: [
+                CSVColumn("Date") { $0.date.formatted() },
+                CSVColumn("Vehicle Name", \.title),
+                CSVColumn("Notes", \.notes)
+            ], 
+            configuration: CSVEncoderConfiguration(dateEncodingStrategy: .iso8601) 
+        )
+        return table.export(rows: events)
+    }
+    
+    func generateCSVFile(vehicle: String) -> URL? {
+        // Get the path to the Documents Directory
+        let fileManager = FileManager.default
+            guard let documentsDirectory = fileManager.urls(
+                for: .documentDirectory, in: .userDomainMask).first else {
+                print("Failed to locate the Documents Directory.")
+                return nil
+            }
+        
+        // Create the file URL
+        let fileName = "\(vehicle)-MaintenanceReport"
+        let fileURL = documentsDirectory.appendingPathComponent(fileName).appendingPathExtension("csv")
+        
+        do {
+                // Save the CSV content to the file
+            try csvData().write(to: fileURL, atomically: true, encoding: .utf8)
+                print("File saved to \(fileURL)")
+                return fileURL
+            } catch {
+                print("Failed to save CSV file: \(error.localizedDescription)")
+                return nil
+            }
+    }
+}
+
+extension CSVFile: Transferable {
+  static var transferRepresentation: some TransferRepresentation {
+    DataRepresentation(exportedContentType: .commaSeparatedText) { file in
+      Data(file.csvData().utf8)
+    }
+  }
+}
+
+internal extension BidirectionalCollection where Element == String {
+    var commaDelimited: String { joined(separator: ",") }
+    var newlineDelimited: String { joined(separator: "\r\n") }
+}
+
+public struct CSVColumn<Record> {
+    /// The header name to use for the column in the CSV file's first row.
+    public private(set) var header: String
+   
+    public private(set) var attribute: (Record) -> CSVEncodable
+    
+    public init(
+        _ header: String,
+        attribute: @escaping (Record) -> CSVEncodable
+    ) {
+        self.header = header
+        self.attribute = attribute
+    }
+}
+
+extension CSVColumn {
+    public init<T: CSVEncodable> (
+        _ header: String,
+        _ keyPath: KeyPath<Record, T>
+    ) {
+        self.init(
+            header,
+            attribute: { $0[keyPath: keyPath] }
+        )
+    }
+}
+
+public protocol CSVEncodable {
+    /// Derive the string representation to be used in the exported CSV.
+    func encode(configuration: CSVEncoderConfiguration) -> String
+}
+
+extension String: CSVEncodable {
+    public func encode(configuration: CSVEncoderConfiguration) -> String {
+        self
+    }
+}
+
+extension Date: CSVEncodable {
+    public func encode(configuration: CSVEncoderConfiguration) -> String {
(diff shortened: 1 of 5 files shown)


mikaelacaron:
awesome work!! I'll look at this PR over the weekend

Basic-Car-Maintenance/Shared/Dashboard/Views/CSVEncoder.swift
```diff
@@ -0,0 +1,210 @@
...
+//  CSVEncoder.swift
+//  Basic-Car-Maintenance
+//
+//  https://github.com/mikaelacaron/Basic-Car-Maintenance
+//  See LICENSE for license information.
+//
+
+import Foundation
+
+internal extension BidirectionalCollection where Element == String {
```
mikaelacaron: this doesn't need to be marked `internal` cause it is by default

Basic-Car-Maintenance/Shared/Dashboard/Views/CSVEncoder.swift
```diff
@@ -0,0 +1,210 @@
...
+    var commaDelimited: String { joined(separator: ",") }
+    var newlineDelimited: String { joined(separator: "\r\n") }
+}
+
+public struct CSVColumn<Record> {
```
mikaelacaron: this shouldn't be `public` the "Shared" folder that this is in, is still all in a single project, so it's fine if it stays just `internal`

Same with like the whole file

Basic-Car-Maintenance/Shared/Dashboard/Views/ExportOptionsView.swift
```diff
@@ -9,12 +9,20 @@
...
+}
+
 struct ExportOptionsView: View {
     @Environment(\.dismiss) var dismiss
     @State private var selectedVehicle: Vehicle?
     @State private var isShowingThumbnail = false
     @State private var pdfDoc: PDFDocument?
     @State private var showingErrorAlert = false
+    @State private var selectedOption: ExportOption?
+    @State private var showingExporter = false
```
mikaelacaron:
```suggestion
    @State private var showingCSVGeneratorView = false
```

renaming this, and how the variable is used is this file, cause I think it feels more clear cause "showExporter" also feels like it would be used for the PDF, but it isn't

Basic-Car-Maintenance/Shared/Dashboard/Views/ExportOptionsView.swift
```diff
@@ -41,22 +49,39 @@ struct ExportOptionsView: View {
             .padding(.horizontal)
             .toolbar {
                 ToolbarItem(placement: .topBarTrailing) {
-                    Button("Export") { 
+                    Menu(content: {
+                        Picker("Export", selection: $selectedOption) {
+                            ForEach(ExportOption.allCases, id: \.self) { option in
```
mikaelacaron:
```suggestion
                            ForEach(ExportOption.allCases) { option in
```

The `id: \.self` isn't needed because `ExportOption` conforms to `Identifiable`

Also for the Menu I use multiple trailing closure syntax, as a personal preference

Basic-Car-Maintenance/Shared/Dashboard/Views/ExportOptionsView.swift
```diff
@@ -41,22 +49,39 @@ struct ExportOptionsView: View {
...
                             if !events.isEmpty {
-                                let pdfGenerator = CarMaintenancePDFGenerator(
-                                    vehicleName: selectedVehicle.name,
-                                    events: events
-                                )
-                                self.pdfDoc = pdfGenerator.generatePDF() 
-                                isShowingThumbnail = true
+                                switch selectedOption {
+                                case .pdf:
+                                    selectedOption = nil
```
mikaelacaron, line 67: I understand why this is being set to nil, but it's not my favorite way to implement this, maybe we could set it to nil on the dismiss of the view that's shown?

but this works! so I'm not too worried about it

Basic-Car-Maintenance/Shared/Dashboard/Views/CSVEncoder.swift
mikaelacaron: also moving this file to just the `Dashboard` folder, cause it doesn't fit being in the `Dashboard` > `Views` folder

Basic-Car-Maintenance/Shared/Dashboard/Views/CSVGeneratorView.swift
```diff
@@ -0,0 +1,118 @@
...
+import Foundation
+import SwiftUI
+
+struct CSVGeneratorView: View {
+    @Environment(\.dismiss) var dismiss
+    
+    let events: [MaintenanceEvent]
+    let vehicleName: String
+    
+    func csvData() -> String {
```
mikaelacaron: moving the functions below the `body`, just as a personal preference thing

and they can be `private`, because they're only used in this view

Basic-Car-Maintenance/Shared/Dashboard/Views/CSVEncoder.swift
```diff
@@ -0,0 +1,210 @@
...
+}
+
+struct CSVColumn<Record> {
+    /// The header name to use for the column in the CSV file's first row.
+    private(set) var header: String
+   
+    private(set) var attribute: (Record) -> CSVEncodable
+    
+    init(
+        _ header: String,
```
mikaelacaron: personal preference, but I don't format this on multiple lines unless it's needed because the line is too long, or there's many parameters (which at that point usually it needs to be refactored for passing too many things

mikaelacaron (approved):
Great work!!! Thanks for contributing!

You added more complexity in the csv than I would've just for exporting the bool in different ways and the date, I would've only done one option, and made it not as generic, but that's fine! It'll be more flexible in the future how you've written it

One small thing, that I'll make a new issue out of later is that the `generateCSVFile` function in `CSVGeneratorView` gets run multiple times, because the function is really called when the view body is evaluated, which can happen many times. A better way to implement this is to make that function trigger from a button tap. But no worries I'll add that as an issue later

PR #372, opened 2024-11-30
feat: Add vehicle-based filtering to DashboardView

### LINKED ISSUE #354: IMPROVE - Add Filter by Vehicle to `DashboardView`
Add a filter to the `DashboardView` to filter by vehicle

This is what the filter currently look likes, and add a new section that lists the vehicle names, and also `All Vehicles`, and depending on which one the user taps, it'll filter the list on the `DashboardView`

Screenshot 2024-10-23 at 2 48 09 PM

N/A, feel free to make suggestions!

Abrahamiaan:
# What it Does
* Closes #354 
* Adds a feature that allows users to filter maintenance events by the selected vehicle

# How I Tested
* Run the application.
* Navigate to the Dashboard screen.
* Tap the Filter button in the toolbar.
* Select By Vehicle from the filter options.
* Choose a specific vehicle from the list and verify that only maintenance events for that vehicle are displayed.
* Repeat the above step, but select All Vehicles to ensure all events reappear.

# Screenshot and Preview

https://github.com/user-attachments/assets/b560d063-748b-418c-8606-9535fc4f9e8a

empty-state

### DIFF
--- a/Basic-Car-Maintenance/Shared/Dashboard/ViewModels/DashboardViewModel.swift
+++ b/Basic-Car-Maintenance/Shared/Dashboard/ViewModels/DashboardViewModel.swift
@@ -23,12 +23,22 @@ class DashboardViewModel {
     var vehicles = [Vehicle]()
     var searchText: String = ""
     var isLoading = false
+    var selectedVehicle: Vehicle?
 
     var sortedEvents: [MaintenanceEvent] {
         switch sortOption {
-        case .oldestToNewest: events.sorted { $0.date < $1.date }
-        case .newestToOldest: events.sorted { $0.date > $1.date }
-        case .custom: events
+        case .oldestToNewest:
+            return events.sorted { $0.date < $1.date }
+        case .newestToOldest:
+            return events.sorted { $0.date > $1.date }
+        case .vehicleType:
+            if let vehicle = selectedVehicle {
+                return events.filter { $0.vehicleID == vehicle.id }
+            } else {
+                return events
+            }
+        case .custom:
+            return events
         }
     }
     
@@ -180,7 +190,8 @@ extension DashboardViewModel {
     enum SortOption: Int, CaseIterable, Identifiable {
         case oldestToNewest = 0
         case newestToOldest = 1
-        case custom = 2
+        case vehicleType = 2
+        case custom = 3
         
         var id: Int {
             rawValue
@@ -200,6 +211,10 @@ extension DashboardViewModel {
                 LocalizedStringResource(
                     "Custom",
                     comment: "Sorting option that sorts items according to the user's preferences.")
+            case .vehicleType:
+                LocalizedStringResource(
+                    "By Vehicle",
+                    comment: "Sorting option that sorts items according to vehicle type.")
             }
         }
     }
--- a/Basic-Car-Maintenance/Shared/Dashboard/ViewModels/DashboardViewModel.swift
+++ b/Basic-Car-Maintenance/Shared/Dashboard/ViewModels/DashboardViewModel.swift
@@ -23,7 +23,7 @@ class DashboardViewModel {
     var vehicles = [Vehicle]()
     var searchText: String = ""
     var isLoading = false
-    var selectedVehicle: Vehicle?
+    var selectedVehicleToSort: Vehicle? 
 
     var sortedEvents: [MaintenanceEvent] {
         switch sortOption {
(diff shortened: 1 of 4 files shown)


Abrhamian:
@mikaelacaron  Hi, have you reviewed the PR? I wonder if there is anything that needs to be improved. Thanks in advance

mikaelacaron:
I'll review this soon! sorry for the delay, thanks for contributing!

Basic-Car-Maintenance/Shared/Dashboard/ViewModels/DashboardViewModel.swift
```diff
@@ -180,7 +190,8 @@ extension DashboardViewModel {
     enum SortOption: Int, CaseIterable, Identifiable {
         case oldestToNewest = 0
         case newestToOldest = 1
-        case custom = 2
+        case vehicleType = 2
```
mikaelacaron: I think `byVehicle` is a fine variable name for this, cause "vehicle type" sounds like we'd separate it by like suv vs truck or something

I'll make these edits

Basic-Car-Maintenance/Shared/Dashboard/ViewModels/DashboardViewModel.swift
```diff
@@ -23,12 +23,22 @@ class DashboardViewModel {
     var vehicles = [Vehicle]()
     var searchText: String = ""
     var isLoading = false
+    var selectedVehicle: Vehicle?
```
mikaelacaron: let's make this more specific like `selectedVehicleToSort` because just `selectedVehicle`, when looking at the `DashboardView` as a whole isn't really clear, and would further cause confusion if later we allow you to see details about some maintenance event

Basic-Car-Maintenance/Shared/Dashboard/Views/DashboardView.swift
```diff
@@ -123,12 +130,21 @@ struct DashboardView: View {
...
                         } label: {
                             EmptyView()
                         }
+                        
+                        Button(action: {
+                            viewModel.sortOption = .vehicleType
+                            isShowingVehicleSelection = true
+                        }, label: {
+                            Text(DashboardViewModel.SortOption.vehicleType.label)
+                        })
```
mikaelacaron: personally I like multiple trailing closure syntax, I make this comment often, do you think I should write this down somewhere?

like this:
```swift
Button {
    // action here
} label: {
}
```

Basic-Car-Maintenance/Shared/Dashboard/Views/VehicleSelectionView.swift
```diff
@@ -0,0 +1,40 @@
...
+//  https://github.com/mikaelacaron/Basic-Car-Maintenance
+//  See LICENSE for license information.
+//
+
+import SwiftUI
+
+struct VehicleSelectionView: View {
+    @Environment(\.dismiss) var dismiss
+    @Binding var selectedVehicle: Vehicle?
+    var vehicles: [Vehicle]
```
mikaelacaron: I order these by scope public is closer to the top, and private is closer to the bottom

so these would go in this order:
```suggestion
    @Binding var selectedVehicle: Vehicle?
    var vehicles: [Vehicle]
    @Environment(\.dismiss) var dismiss
```

usually I put the binding first and then whatever variable, and Environment (cause it's not passed in next

Basic-Car-Maintenance/Shared/Dashboard/Views/VehicleSelectionView.swift
```diff
@@ -0,0 +1,40 @@
...
+                Button(action: {
+                    selectedVehicle = nil
+                    dismiss()
+                }, label: {
+                    Text("All Vehicles")
+                        .fontWeight(.bold)
+                        .foregroundColor(.blue)
+                })
+                
+                ForEach(vehicles, id: \.id) { vehicle in
```
mikaelacaron: you don't need `id: \.id` because Vehicle is `Identifiable`

```suggestion
                ForEach(vehicles) { vehicle in
```

mikaelacaron (approved):
Great work!! I would've preferred to keep the functionality just in popup, but honestly this is totally fine and it works, thank you!

I made comments on little thing, mainly formatting, and fixed those, so no worries! thanks again for contributing!

PR #376, opened 2025-02-10
Implemented "Rate App" Button in `SettingsView`

### LINKED ISSUE #304: FEATURE - Add Rate App Button
Add a new button in the `SettingsView` that says "Rate the App" when tapped it opens the popup in the App Store to rate the app

Add this under the "Report a bug" row

* N/A, there's an API to do this, more info here https://www.avanderlee.com/swift/skstorereviewcontroller-app-ratings/

## Additional context
N/A

paapa-kusi:
# What it Does
* Closes #304 
* Creates a Rate App option in the settings view of the application.

# How I Tested
* Used the preview and a real iPhone to test the implementation, the option works on both.

# Notes
* Used the API in the description of the issue for help with implementation.

# Screenshot
*
Screenshot 2025-02-10 at 12 21 48 PM

### DIFF
--- a/Basic-Car-Maintenance/Shared/Settings/Views/SettingsView.swift
+++ b/Basic-Car-Maintenance/Shared/Settings/Views/SettingsView.swift
@@ -11,8 +11,6 @@ import UniformTypeIdentifiers
 import TipKit
 import StoreKit
 
-
-
 struct SettingsView: View {
     @Environment(ActionService.self) var actionService
     @Environment(\.scenePhase) var scenePhase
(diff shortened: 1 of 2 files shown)


Basic-Car-Maintenance/Shared/Settings/Views/SettingsView.swift
```diff
@@ -82,6 +85,18 @@ struct SettingsView: View {
                         }
                     }
                     
+                    // rate app button
```
mikaelacaron: No need to add this comment

Basic-Car-Maintenance/Shared/Settings/Views/SettingsView.swift
```diff
@@ -82,6 +85,18 @@ struct SettingsView: View {
...
+                    Button( action:{requestAppReview()} ) {
```
mikaelacaron: Can you make this multiple trailing closure syntax like

```swift
Button {
   // action here
} label: {
   // Label
}
```

Basic-Car-Maintenance/Shared/Settings/Views/SettingsView.swift
```diff
@@ -312,7 +327,18 @@ struct SettingsView: View {
             }
         }
     }
-}
+    private func requestAppReview() {
+            if let scene = UIApplication.shared.connectedScenes.first as? UIWindowScene {
+                if #available(iOS 18.0, *) {
+                    // Use AppStore.requestReview for iOS 18+
+                    AppStore.requestReview(in: scene)
```
mikaelacaron: You don't need the fallback, I'll be making this iOS 18 only

mikaelacaron (changes requested):
Great work!! Just a few small comments on formatting

mikaelacaron:
Please resolve this and then select request review again

paapa-kusi (commented):
I've just submitted those changes, thank you for the feedback! Please let me know if any other changes are needed.

Basic-Car-Maintenance/Shared/Settings/Views/SettingsView.swift
```diff
@@ -9,12 +9,15 @@
 import SwiftUI
 import UniformTypeIdentifiers
 import TipKit
+import StoreKit
+
+
```
mikaelacaron: SwiftLint should've shown a warning about this, because there's no need for the extra line breaks

Basic-Car-Maintenance/Shared/Settings/Views/SettingsView.swift
```diff
@@ -208,26 +224,26 @@ struct SettingsView: View {
...
-                        // A toast view to notify the user of version copy
-                        Text("Copied!", comment: "Text to notify user that app version was copied")
-                            .font(.callout)
-                            .padding(8)
-                            .foregroundStyle(colorScheme == .light ? .white : .black)
-                            .background(colorScheme == .light ? .black : .white)
-                            .clipShape(Capsule())
-                            .opacity(copiedAppVersion ? 1 : 0)
-                            .animation(.linear(duration: 0.2), value: copiedAppVersion)
+                .frame(maxWidth: .infinity, alignment: .center)
```
mikaelacaron: This looks like a bunch of whitespace change, next time don't add it into this PR but make an issue and new PR for something like this, cause these 20 lines aren't really related to this issue.

and if this was a bigger PR and hundreds of lines of white space were changed, I wouldn't be able to tell if something is actually important or not, or if it was whitespace

Basic-Car-Maintenance/Shared/Settings/Views/SettingsView.swift
```diff
@@ -312,6 +328,11 @@ struct SettingsView: View {
             }
         }
     }
+    private func requestAppReview() {
+        if let scene = UIApplication.shared.connectedScenes.first as? UIWindowScene {
+            AppStore.requestReview(in: scene)
```
mikaelacaron: this specific function says:

> Because this method may not present an alert, don’t call `requestReview` or `requestReview(in:)` in response to a button tap or other user action.

[in the docs](https://developer.apple.com/documentation/storekit/appstore/3954432-requestreview), which makes me think this isn't the function to use

[This doc](https://developer.apple.com/documentation/storekit/requesting-app-store-reviews#Manually-request-a-review) details it this way instead, as a response to a button press

mikaelacaron (approved):
Great work!! Thanks for contributing! I forgot to mention, but when I make comments on a PR, click the resolve button on the comments that you address, cause then I can see that you addressed those issues.

I used a different function for the rating review, see the comment about why

And after seeing that, I decided to put this behind a flag, because we don't have all the info yet to make the URL for showing the app store product page

mikaelacaron:
also note that the `Localizable.xcstrings` file was added, which means you maybe didn't build the project??

This should have a change, that happens automatically when building the project, because you added a new string.

PR #378, opened 2025-02-20
Improve README - project description, features, tech stack

radliff:
This pull request makes the readME clearer & more informative to first-time users and contributors by adding a description of the project and what it's made for. Additionally, it introduces a feature list so anyone who wants to know exactly what the app does at first glance can do so.  It also adds a small call to action at the end.

### DIFF
--- a/README.md
+++ b/README.md
@@ -16,12 +16,19 @@ It's made so users can efficiently track & manage their vehicle's maintenance sc
 
 ## Features
 **Maintenance Tracking** - Log & monitor oil changes, tire rotations, brake checks, etc.
+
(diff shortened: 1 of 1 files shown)


mikaelacaron (approved):
Thanks for adding this!  It's a great addition to help others know what's in the app when they contribute. Be sure you actually know what's available. Half of the features you listed are not in the app. Which leads me to believe it's AI written, which is fine, but make sure you're accurate.

I changed some of the formatting

In the PR, please keep the original template that's there, even for a change like this, which would've noted that no Swift code was changed, but only the README file

PR #381, opened 2025-02-28
Added a filter in OdometerView 

### LINKED ISSUE #350: IMPROVE - Filter `OdometerView` by Vehicle
* Add a filter button to the toolbar, to the left of the `+` button on the `OdometerView` and then show a context menu to filter the vehicles shown
   * See the same position of the button as the `DashboardView`
* There should be options like "Vehicle name 1", "vehicle name 2", "All Vehicles", so it's showing the vehicle names, and then by default `All Vehicles` will be selected
* tapping on the vehicle name will filter the `readings` shown in the `OdometerView`

N/A

SanjaySiddharth:
# What it Does
* Closes #350 
* Added a filter in OdometerView which shows the readings of either all the vehicles(default) or any one selected vehicle
* Fixed typo in a commented line
* Added 3rd car and another reading to the preview in OdometerView

# How I Tested
* Run the application
* Add three vehicles from the Settings tab
* Navigate to Odometer screen and create some readings for each vehicle
* Click on the filter button in the toolbar (prior to the plus sign(+))
* Choose the vehicle you want to filter on . By default it shows all the vehicles.
* Repeat the above step, but select All Vehicles to ensure all readings reappear.

# Screenshot


mikaelacaron:
Thank for the contribution! I'll try to review this soon

ClarkLiam (approved):

Basic-Car-Maintenance/Shared/Odometer/ViewModels/OdometerViewModel.swift
```diff
@@ -22,8 +22,9 @@ class OdometerViewModel {
     var showEditErrorAlert = false
     var selectedReading: OdometerReading?
     var isShowingEditReadingView = false
-    
+    var filterOption: String = "No vehicles found"
```
mikaelacaron: This variable isn't used anywhere, so let's remove it

Basic-Car-Maintenance/Shared/Odometer/Views/OdometerView.swift
```diff
@@ -14,6 +14,8 @@ struct OdometerView: View {
     
     @State private var viewModel: OdometerViewModel
     @State private var selectedTimeRange: TimeRange = .all
+    // start of my variables
+    @State private var selectedVehicleID: String?
```
mikaelacaron: no need for this comment also a variable for `selectedVehicleID` was created in the `OdometerViewModel` so it should be used, rather than making a new variable in the view, which would lead to two different sources of truth

And this is a string, which makes sense for specifically keeping track of the vehicleID, but then later setting something to `= "nil"` doesn't feel right cause that does have a value and feels like a code smell

Basic-Car-Maintenance/Shared/Odometer/Views/OdometerView.swift
```diff
@@ -99,6 +101,29 @@ struct OdometerView: View {
             }
             .toolbar {
                 ToolbarItemGroup(placement: .primaryAction) {
+                    Menu {
+                        Button("All") {
+                            selectedVehicleID = "nil"
+                            applyFilter()
+                            print(selectedVehicleID!)
+                        }
+                        ForEach(viewModel.vehicles) { value in
```
mikaelacaron: Instead of using the variable name `value`, use `vehicle` because it's more descriptive

Basic-Car-Maintenance/Shared/Odometer/Views/OdometerView.swift
```diff
@@ -128,7 +153,32 @@ struct OdometerView: View {
     }
     
     // swiftlint:disable:next line_length
-    /// Filter the readins based on the selected time range, and if there are no readings in the last 30 days, just show the last reading.
+    /// Filter the readings based on the selected vehicle , and if no vehicle is selected , it shows all the vehicle's readings
+    /// - Parameter : None
+    /// - Returns: The `[OdometerReading]`s for all the vehicles / only the filtered vehicle
+    /// 
+    private func filteredReadings() -> [OdometerReading] {
```
mikaelacaron: this can actually be a computed property as opposed to a function, because there are no inputs and it reacts to variable changes in the viewModel

mikaelacaron (approved):
Thank you so much for contributing!! I appreciate your efforts and patience for this super delayed review and merge

Your implementation works great, but I left a few comments mostly around just how the implementation
Thanks again!

mikaelacaron:
one tip, when making a branch name, the name `350-improve-filter-by-vehicle` would've been best. using `#` and `( )` in the branch name isn't conventional

PR #334, opened 2024-10-08
50 medium widget

aking618:
# What it Does
* Closes #50
* This PR adds a medium widget for viewing a vehicle's latest maintenance events logged.
    *  Due to widget sizing, only up to two events are visible. (The most current events are used)

# How I Tested
* Run the application
* Add a vehicle / maintenance 
* Return to the home screen
* Add a widget
* Configure the widget. (Ensure that all vehicle names are present)

# Notes
* I added an app group to share user defaults with the widget. I currently store the user id there to use it for fetching vehicles.
    * I originally attempted to use `FirebaseAuth`, but I couldn't get anything to work. 

# Screenshots
| Widget Gallery | View |
| --- | --- |
|   |  |

I would love to hear some feedback and ideas on the data sharing.

PR #394, opened 2025-10-09
Feat: add small total maintenance events count widget

### LINKED ISSUE #15: FEATURE - Small Widget, View Maintenance Events
* Create a small widget that allows you to view the total number of maintenance events for a specific vehicle
   * Be able to configure which vehicle you want to view for this widget

* N/A

## Additional context
* Only make the small widget (I'll create new issues for adding different widget sizes later

TMaszko:
# What it Does
* Closes #15 
* It adds small widget that displays total maintenance events count for a selected vehicle. User can configure selected vehicle by long pressing on it.

# How I Tested
- Run the **Widget schema**
- Add a new vehicle in the main app
- Check the Home Screen -> Now widget will look like so:
Screenshot 2025-10-09 at 23 29 20

- Long press the widget and select the vehicle
image

- Select a vehicle
image

- In case of no events you will see something like this:
image

- In case of any events:
image

- In case of an unexpected error:
Screenshot 2025-10-10 at 00 08 56

# Notes
* Cross sharing Firebase authentication [here](https://firebase.google.com/docs/auth/ios/single-sign-on)
* Kudos to [aking618](https://github.com/mikaelacaron/Basic-Car-Maintenance/pull/334) for a great work that this PR is **heavily** inspired by.
* Designs are not final ;) I couldn't quickly find any better solution though :D
* This is my first PR in Swift so please keep that in mind ;)

# Screenshot
Screenshot from the widget gallery
image


Basic-Car-Maintenance-Widget/BasicCarMaintenanceWidget.swift
```diff
@@ -5,57 +5,67 @@
...
+        var entries: [MaintenanceEventsCountEntry] = []
+        
         let currentDate = Date()
-        for hourOffset in 0 ..< 5 {
-            let entryDate = Calendar.current.date(byAdding: .hour, value: hourOffset, to: currentDate)!
-            let entry = SimpleEntry(date: entryDate, configuration: configuration)
-            entries.append(entry)
+        let nextUpdate = Calendar.current.date(byAdding: .hour, value: 1, to: currentDate)!
+        let vehicleID = configuration.selectedVehicle?.id;
+        let result = await DataService.fetchMaintenanceEventsCount(for: vehicleID)
```
mikaelacaron: instead of `DataService.fetchMaintenanceEventsCount` returning the `Result` type instead let's just use an async throws kind of function, so then it's handled using do-catch

Basic-Car-Maintenance-Widget/CarInfoWithMaintenanceEventsCount.swift
```diff
@@ -0,0 +1,39 @@
...
+//
+
+import SwiftUI
+
+struct CarInfoWithMaintenanceEventsCount: View {
+    let vehicle: VehicleAppEntity
+    let maintenanceEventsCount: Int
+    
+    var body: some View {
+        VStack(alignment: .leading) {
```
mikaelacaron, line 16: throughout this view I added line breaks because I like the modifiers to be on their own lines

Basic-Car-Maintenance-Widget/VehicleAppEntity.swift
```diff
@@ -0,0 +1,49 @@
...
+        self.displayString = displayString
+        self.data = data
+    }
+}
+
+struct VehicleAppEntityQuery: EntityQuery {
+    func entities(
+        for identifiers: [VehicleAppEntity.ID]
+    ) async throws -> [VehicleAppEntity] {
+        let result = await DataService.fetchVehicles()
```
mikaelacaron: Same here about using Result type, but I'll comment that in the `DataService` too

Basic-Car-Maintenance/Shared/Settings/ViewModels/AuthenticationViewModel.swift
```diff
@@ -47,6 +47,7 @@ final class AuthenticationViewModel {
     
     func signInAnonymously() {
         Task {
+            // TODO: remove that!!
```
mikaelacaron: Remove them for what reason?

Basic-Car-Maintenance/Shared/BasicCarMaintenanceApp.swift
```diff
@@ -45,8 +45,9 @@ class AppDelegate: NSObject, UIApplicationDelegate {
     ) -> Bool {
         
         FirebaseApp.configure()
-        
-        let useEmulator = UserDefaults.standard.bool(forKey: "useEmulator")
+    
+
+        let useEmulator = true
```
mikaelacaron, line 49: I'm fine with this change because for this project everyone should be using the emulator, but in other projects, this may be something that needs justified on why it's been changed

Basic-Car-Maintenance/Shared/GoogleService-Info.plist
```diff
@@ -19,16 +19,16 @@
...
+	<true/>
 	<key>IS_SIGNIN_ENABLED</key>
-	<true></true>
+	<true/>
 	<key>GOOGLE_APP_ID</key>
 	<string>1:255838601762:ios:12417d95fcb477338f807b</string>
 </dict>
-</plist>
\ No newline at end of file
+</plist>
```
mikaelacaron, line 34: why does this need changed?

Basic-Car-Maintenance/Shared/Info.plist
```diff
@@ -2,10 +2,12 @@
 <!DOCTYPE plist PUBLIC "-//Apple//DTD PLIST 1.0//EN" "http://www.apple.com/DTDs/PropertyList-1.0.dtd">
 <plist version="1.0">
 <dict>
-	<key>FirebaseAutomaticScreenReportingEnabled</key>
-	<false/>
+	<key>KeychainAccessGroup</key>
+	<string>$(AppIdentifierPrefix)com.mycompany.Basic-Car-Maintenance.Shared</string>
```
mikaelacaron: This looks hard coded, and instead we'd need a way to properly use the xcconfig file, using the bundle ID, `PRODUCT_BUNDLE_IDENTIFIER` because it should be unique for each person who's contributed to the project

If you're not sure how to fix this, drop a comment and I can do it before merging this

Basic-Car-Maintenance/Basic_Car_Maintenance.entitlements
```diff
@@ -10,5 +10,9 @@
 	<true/>
 	<key>com.apple.security.files.user-selected.read-only</key>
 	<true/>
+	<key>keychain-access-groups</key>
+	<array>
+		<string>$(AppIdentifierPrefix)com.mycompany.Basic-Car-Maintenance.Shared</string>
```
mikaelacaron: same this this using the xcconfig file

Basic-Car-Maintenance-WidgetExtension.entitlements
```diff
@@ -0,0 +1,10 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<!DOCTYPE plist PUBLIC "-//Apple//DTD PLIST 1.0//EN" "http://www.apple.com/DTDs/PropertyList-1.0.dtd">
+<plist version="1.0">
+<dict>
+	<key>keychain-access-groups</key>
+	<array>
+		<string>$(AppIdentifierPrefix)com.mycompany.Basic-Car-Maintenance.Shared</string>
```
mikaelacaron: same here about the xcconfig file

Basic-Car-Maintenance.xcodeproj/project.pbxproj
```diff
@@ -632,7 +664,7 @@
 				"INFOPLIST_KEY_UIStatusBarStyle[sdk=iphonesimulator*]" = UIStatusBarStyleDefault;
 				INFOPLIST_KEY_UISupportedInterfaceOrientations_iPad = "UIInterfaceOrientationPortrait UIInterfaceOrientationPortraitUpsideDown UIInterfaceOrientationLandscapeLeft UIInterfaceOrientationLandscapeRight";
 				INFOPLIST_KEY_UISupportedInterfaceOrientations_iPhone = "UIInterfaceOrientationPortrait UIInterfaceOrientationLandscapeLeft UIInterfaceOrientationLandscapeRight";
-				IPHONEOS_DEPLOYMENT_TARGET = 26.0;
+				IPHONEOS_DEPLOYMENT_TARGET = 18.0;
```
mikaelacaron: this shouldn't be changed back to 18, it should be set to iOS26 now. I think I fixed this

Basic-Car-Maintenance-Widget/DataService.swift
```diff
@@ -0,0 +1,103 @@
...
+//  Basic-Car-Maintenance
+//
+//  https://github.com/mikaelacaron/Basic-Car-Maintenance
+//  See LICENSE for license information.
+//
+
+import FirebaseAuth
+import FirebaseFirestore
+
+enum DataService {
```
mikaelacaron, line 12: Please change these functions and then the call sites to use async throws rather than using Result type

so it would be some like this:
```swift
func fetchMaintenanceEventsCount(for vehicleID: String?) async throws -> Int {
}


// call site:
do {
   let count = fetchMaintenanceEventsCount("idHere")
} catch {
   // handle error
}
```

mikaelacaron (changes requested):
Great work!! Main thing to change is the `DataService` to use async throws, rather than the `Result` type

I've also updated various comments / line breaks here and there, so be sure to pull changes, and then make the last change that I noted! Thank you, then I'll re-review

If you have any questions, you can comment directly on that thread, otherwise click the resolve button once you've addressed my comment. Once you fix everything, please click the re-review button!

Image

Basic-Car-Maintenance/Shared/GoogleService-Info.plist
TMaszko: Xcode has done it automatically :/

Basic-Car-Maintenance/Shared/Settings/ViewModels/AuthenticationViewModel.swift
TMaszko: I had `signInWithEmail` for testing locally purpose ;) so it's just a leftover ;)

Basic-Car-Maintenance/Shared/BasicCarMaintenanceApp.swift
TMaszko: Yeah ... Command line arguments doesn't work for a widget unfortunately since it's a separate process. I was thinking about using environment variables but I'm not sure if they would work.

Basic-Car-Maintenance.xcodeproj/project.pbxproj
TMaszko: My fault yeah. I haven't updated Xcode yet so I had to do that to be able to even start working on this 😂

Basic-Car-Maintenance/Shared/Info.plist
TMaszko: Since it's a pure config stuff ;) I'll leave it for you

mikaelacaron (approved):
Great work!! 👏  thank you!!

It wasn't showing the widget when I was signed in anonymously, but I'll Sign in with Apple and see if that changes, but otherwise that's totally fine! We can always fix that later if needed

PR #415, opened 2026-07-30
Fixed odometer reading to update properly

### LINKED ISSUE #297: BUG - Odometer Reading Doesn't Update Properly
## Describe the bug
* When editing the odometer reading, it doesn't update on the `OdometerView` after the `EditOdometerReadingView` is dismissed

## To Reproduce
1. Make sure you've added a vehicle
2. Add an odometer reading
3. Swipe left, and tap edit
4. Edit the odometer reading amount
5. Tap Update, and the view dismisses

## Expected behavior
* The `OdometerView` should see the updated value, but it's still showing the old value

## Screenshots

https://github.com/user-attachments/assets/fb4c7e5b-05b4-46c2-a72b-b92bcc7c02a8

 * OS: iOS 17.5.1
 * App Version v1.0

## Additional Context
* If you can do it without another network call that would be great, but if not, that's fine

ShaniDan:
# What it Does
* Closes #297 
* Fixes the issue where the odometer reading was not updating properly by including the missing `userID` when updating the odometer record.

# How I Tested
* Ran the application
* Verified that the userID is present by putting breakpoint and debugging it with po in the debugger

# Notes
* 

# Screenshot
* https://github.com/user-attachments/assets/b62b4f23-a5ab-4bf8-97f8-e7994acfbd0a

### DIFF
--- a/Basic-Car-Maintenance/Shared/Odometer/ViewModels/OdometerViewModel.swift
+++ b/Basic-Car-Maintenance/Shared/Odometer/ViewModels/OdometerViewModel.swift
@@ -58,15 +58,19 @@ class OdometerViewModel {
     }
     
     func updateOdometerReading(_ reading: OdometerReading) {
-        do {
-            try firebaseService.updateReading(reading)
-            
-            AnalyticsService.shared.logEvent(.odometerUpdate)
-            
-            isShowingEditReadingView = false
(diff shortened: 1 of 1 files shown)


mikaelacaron (approved):
Amazing work!! 🥳

